PR #12, opened 2019-01-10
Added configuration for Munin

matq007:
- [x] Added `munin.config`
- [x] Updated docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Currently documentation is available for the following clusters:
 * [GIS](docs/gis.md)
 * [HEBBE](docs/hebbe.md)
 * [MENDEL](docs/mendel.md)
+* [MUNIN](docs/munin.md)
 * [PHOENIX](docs/phoenix.md)
 * [SHH](docs/shh.md)
 * [UCT_HEX](docs/uct_hex.md)
--- a/conf/munin.config
+++ b/conf/munin.config
@@ -10,10 +10,15 @@ process {
   executor = 'local'
 }
 
+sngularity {
+  enabled = true
+  autoMounts = true
+}
+
 params {
   saveReference = true
 
-  max_memory = 256.GB
+  max_memory = 128.GB
   max_cpus = 16
   max_time = 72.h
   // illumina iGenomes reference file paths on UPPMAX
--- a/conf/munin.config
+++ b/conf/munin.config
@@ -1,0 +1,21 @@
+//Profile config names for nf-core/configs
+params {
+  config_profile_name = 'MUNIN'
+  config_profile_description = 'Big iron cluster profile provided by nf-core/configs.'
+  config_profile_contact = 'Szilveszter Juhos (@szilva)'
+  config_profile_url = ''
+}
+
+process {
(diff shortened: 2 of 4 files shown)


apeltzer:
Yes, these should be possible to keep them in the `munin.config` and then document that users would need to run it accordingly:

`nextflow run -profile singularity,munin`
or 
`nextflow run -profile docker,munin` 

👍

ewels:
I think it makes most sense if Munin always runs with _either_ Docker _or_ Singularity. Then we can just put that in the config and remove the need for any customisation from the user. It's not difficult to override this if necessary.

matq007:
Munin is capable of running both Singularity and Docker. With the horsepower it possesses I think it makes sense to keep both methods. The additional info is because of the permissions. I can specify that in order to run docker run `nextflow run -profile docker,munin` and for singularity run `nextflow run -profile docker,munin` as @apeltzer proposed.

ewels:
Basically, I fully expect users to accidentally run `nextflow run -profile munin` and then wonder why it doesn't work. It's usually good to reduce as much as possible the number of command-line flags that are required for execution (my mantra: _sensible defaults_)

apeltzer:
I agree with Phil here - also it's not really a difference whether users use Singularity or Docker, not even computationally.

matq007:
Done. I removed the docker doc also just in case.

drpatelh:
> 
> 
> Done. I removed the docker doc also just in case.

The changes made dont accomodate the configuration for docker at all now. I think what @ewels meant was to put this in the config too:

```
docker {
  mountFlags = 'z'
  fixOwnership = true
}
```

and then to provide a line in the docs suggesting that you can run docker too if required with:
`nextflow run -profile docker,munin`. 

However, as you have rightly mentioned in the docs Singularity will be run by default.

ewels:
Yup, though note that the order of profiles is important. So it needs to be `nextflow run -profile munin,docker` instead of `nextflow run -profile docker,munin`

ewels:
You can even put `enabled:false` into the config along with a comment to make it clear to anyone reading that it's not the default...

```nextflow
// To use docker instead of singularity, use nextflow run -profile munin,docker
docker {
  enabled = false
  mountFlags = 'z'
  fixOwnership = true
}
```

drpatelh:
This is in the template and some of the other docs:
`Using this profile, Nextflow will download a temporary singularity image with all of the required software before execution of the pipeline.`

Strictly speaking we arent doing this anymore right? We are now downloading a docker image which is being converted to a Singularity image. Is this more appropriate:
`Using this profile, a temporary docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.`

I can make these changes in a separate PR.

drpatelh:
@matq007 Looks good to me now :+1:

apeltzer:
Yeah you're right @drpatelh - it would be nice to be more explanatory in the template for this!

drpatelh:
Addressed in PR:
https://github.com/nf-core/configs/pull/13

PR #14, opened 2019-01-21
Made small SLURM queue change for SHH

jfy133:


jfy133:
Merging myself as extremely minor MPI-SHH specific change.

PR #16, opened 2019-02-27
update config for MENDEL cluster

phue:


conf/mendel.config
```diff
@@ -2,21 +2,23 @@
...
   config_profile_name = 'MENDEL'
   config_profile_description = 'GMI MENDEL cluster profile provided by nf-core/configs'
-  config_profile_contact = 'Philipp H (@phue)'
+  config_profile_contact = 'Patrick Hüther (@phue)'
   config_profile_url = 'http://www.gmi.oeaw.ac.at/'
 }
 
-singularity {
-  enabled = true
+manifest {
```
apeltzer: Is this functional code? I'm not so sure about this unfortunately...

conf/mendel.config
```diff
@@ -2,21 +2,23 @@
 params {
   config_profile_name = 'MENDEL'
   config_profile_description = 'GMI MENDEL cluster profile provided by nf-core/configs'
-  config_profile_contact = 'Philipp H (@phue)'
+  config_profile_contact = 'Patrick Hüther (@phue)'
```
apeltzer: I guess that was my fault  - sorry about that!

conf/mendel.config
```diff
@@ -2,21 +2,23 @@
...
-singularity {
-  enabled = true
+manifest {
+  nextflowVersion = '>=19.01.0'
 }
 
 process {
-  beforeScript = 'module load Singularity'
-  executor = 'pbs'
+  beforeScript = {'module load Singularity; module load Miniconda3'}
```
apeltzer: Is there a reason why you a.) load both and b.) not an explicit version of the tool?

conf/mendel.config
phue: No worries :)

conf/mendel.config
phue: a) loading both leaves more options, our Singularity deployment has issues and most users will use conda. However, some people (including me) like containers.
b) In our environment, the latest version is loaded by default. I know this is not best practice but these things rarely get updated here and never caused breakage so far

conf/mendel.config
phue: Surprisingly yes, it seems to work. It overrides the required nextflow version specified in the pipeline, which is needed here because the pbspro scheduler was only recently added.

phue:
> Some minor changes. Also, we require documentation for all cluster configuration profiles - can you add that in the `docs` folder? There is also a template for you to adapt there...

done!

docs/mendel.md
```diff
@@ -0,0 +1,30 @@
...
+
+To use, run the pipeline with `-profile conda,mendel`. This will download and launch the [`mendel.config`](../conf/mendel.config) which has been pre-configured with a setup suitable for the MENDEL cluster. A Conda environment will be created automatically and software dependencies will be downloaded from ['bioconda'](https://bioconda.github.io/).
+
+Theoretically, using `-profile singularity,mendel` would download a docker image containing all of the required software, and convert it to a Singularity image before execution of the pipeline. However, there is a regression in the Singularity deployment on MENDEL which renders containers downloaded from public repositories unusable because they lack the /lustre mountpoint.
+
+If you want to run the pipeline containerized anyway you will have to build the image yourself (on a machine where you have root access) using the provided `Singularity` file in the pipeline repository:
+
+```bash
+cd /path/to/pipeline-repository
+echo 'mkdir /lustre > Singularity'
```
ewels: I think this will overwrite the existing `Singularity` file, no? You presumably want `>>` instead..

docs/mendel.md
```diff
@@ -0,0 +1,30 @@
+# nf-core/configs: MENDEL Configuration
+
+All nf-core pipelines have been successfully configured for use on the MENDEL CLUSTER at the Gregor Mendel Institute (GMI).
+
+To use, run the pipeline with `-profile conda,mendel`. This will download and launch the [`mendel.config`](../conf/mendel.config) which has been pre-configured with a setup suitable for the MENDEL cluster. A Conda environment will be created automatically and software dependencies will be downloaded from ['bioconda'](https://bioconda.github.io/).
+
+Theoretically, using `-profile singularity,mendel` would download a docker image containing all of the required software, and convert it to a Singularity image before execution of the pipeline. However, there is a regression in the Singularity deployment on MENDEL which renders containers downloaded from public repositories unusable because they lack the /lustre mountpoint.
+
+If you want to run the pipeline containerized anyway you will have to build the image yourself (on a machine where you have root access) using the provided `Singularity` file in the pipeline repository:
```
ewels: Note that we are systematically removing all of these files from pipelines now for simplicity (they were originally added for use with singularity-hub, which we are now not pursuing).

docs/mendel.md
```diff
@@ -0,0 +1,30 @@
+# nf-core/configs: MENDEL Configuration
+
+All nf-core pipelines have been successfully configured for use on the MENDEL CLUSTER at the Gregor Mendel Institute (GMI).
+
+To use, run the pipeline with `-profile conda,mendel`. This will download and launch the [`mendel.config`](../conf/mendel.config) which has been pre-configured with a setup suitable for the MENDEL cluster. A Conda environment will be created automatically and software dependencies will be downloaded from ['bioconda'](https://bioconda.github.io/).
+
+Theoretically, using `-profile singularity,mendel` would download a docker image containing all of the required software, and convert it to a Singularity image before execution of the pipeline. However, there is a regression in the Singularity deployment on MENDEL which renders containers downloaded from public repositories unusable because they lack the /lustre mountpoint.
```
ewels: Have you contacted your sysadmins about this? A number of users have encountered the same problem, but it's usually fixable by changing some configuration options with the singularity installation..

docs/mendel.md
```diff
@@ -0,0 +1,30 @@
...
+
+If you want to run the pipeline containerized anyway you will have to build the image yourself (on a machine where you have root access) using the provided `Singularity` file in the pipeline repository:
+
+```bash
+cd /path/to/pipeline-repository
+echo 'mkdir /lustre > Singularity'
+singularity build nf-core-methylseq-custom.simg Singularity
+```
+
+After you copied the container image to the cluster filesystem, make sure to pass the path to the image to the pipeline with `-with-singularity /path/to/nf-core-methylseq-custom.simg`
```
ewels: Minor point, but if using `-with-singularity` then there is no need for the `singularity` profile.

docs/mendel.md
```diff
@@ -0,0 +1,30 @@
...
+
+Before running the pipeline you will need to load Nextflow and Conda using the environment module system on MENDEL. You can do this by issuing the commands below:
+
+```bash
+## Load Nextflow and Conda environment modules
+module purge
+module load Nextflow
+module load Miniconda3 # not needed if using Singularity
```
ewels: Is this not done automatically using your new nextflow config file?

docs/mendel.md
drpatelh:
```
singularity {
  enabled = true
  autoMounts = true
}
```
The `autoMounts` switch worked for us.

drpatelh: Would make the docs and config much simpler if you insist that Singularity has to work, or find a solution to the mount problem. Just avoid version 3.0.0!

https://github.com/nextflow-io/nextflow/issues/897

docs/mendel.md
phue: oops, Good catch! Thanks.

docs/mendel.md
phue: Thanks for the suggestions! 
Our compute nodes run a rather dated linux kernel which doesn't support overlayfs, that's why it doesn't work. Atleast that's what sysadmins keep telling me. I'll dig a bit deeper into this

docs/mendel.md
phue: Though Miniconda is loaded for process execution on the compute nodes according to the config file, it needs to be loaded manually on the login node in order to create the environment on the first run

phue:
Thanks for your comments, I'll adress the issues and submit a new PR.

docs/mendel.md
phue: That is unfortunate. This means I probably should remove the singularity parts entirely from the config and documentation and just support conda

ewels: Nope, you misunderstand me. We're not dropping support for singularity (most people involved in the project use it as their primary environment), we're dropping support for singularity _hub_ - that is, the singularity container registry: https://singularity-hub.org/

We still have support for using singularity, we just get the containers by building from dockerhub instead. We only ever put a couple of pipelines on singularity hub and it never worked properly.

ewels:
Thanks!

docs/mendel.md
drpatelh: I thought there was a misunderstanding too but it seems like the Docker -> Singularity conversion doesnt work for @phue . The Singularity images were being built manually from the Singularity file which will be taken out of most pipelines.

phue: Exactly, pulling Images from the registry doesn't work for us because our Kernel lacks overlayfs (and I reconfirmed with sysadmins, they won't fix it because the current cluster will go out of service soonish) . I have to build the containers manually. I really think t's best to just remove singularity support here until the new cluster is running and rely in conda in the meantime

ewels: Oh sorry, my bad.

I think there _should_ be a way to build “child” containers where overlayfs isn’t an option. I feel like I remember thinking about this a while ago.

Basically have a mini recipe where you use `FROM` (or the singularity equivalent) to point to the nf-core image, then have a bunch of `mkdir` commands. Does that make sense?

phue: >  Does that make sense?

Not sure if I'm understanding this correctly. Wouldn't this also require some sort of recipe which again needs to be built manually?

The only way I see this working currently is either a) to make the docker image somehow writeable (not sure if that's even possible) and then use some sort of hack to add the missing mountpoint using something like `singularity.runOptions = 'mkdir /lustre &&'` or b) to change the dockerfiles to create the folder at buildtime.

But both would require global changes just to accommodate a bug in our institutes' infrastructure. I don't think we should start doing such things, right?

ewels: I've never been able to test solutions to this as we don't have access to a server with overlayfs turned off. But I think that this one-off manual step for the singularity containers should work?

```bash
singularity pull --name nf-core-rnaseq-1.1.simg docker://nfcore/rnaseq:1.1
sudo singularity exec --writable nf-core-rnaseq-1.1.simg mkdir /lustre
```

See http://singularity.lbl.gov/archive/docs/v2-3/docs-changing-containers

Or my child container suggestion was to make your own singularity build script for each pipeline which would look like this:

```Singularity
Bootstrap: docker
From: nfcore/rnaseq:1.1

%post
    mkdir /lustre
```

I have a vague memory of @remiolsen having some kind of nicer run-time nextflow config option for binding directories in Singularity for non-overlayfs people too. But I can't remember the details, or whether it worked. Basically in your nextflow config file having this I think:
```nextflow
singularity.engineOptions = '-B /lustre:/lustre'
```

PR #19, opened 2019-03-06
Provide fix for online issues

apeltzer:


nfcore_custom.config
```diff
@@ -8,7 +8,7 @@
  * name here.
  */
 
-config_base = "${params.custom_config_base}/conf"
+config_base = params.custom_config_base ?: "${params.custom_config_base}/conf" : "https://raw.githubusercontent.com/nf-core/configs/${params.custom_config_version}/conf"
```
sven1103:
```suggestion
default_uri = "https://raw.githubusercontent.com/nf-core/configs/master/conf"
config_base = "${params.custom_config_base ?: default_uri}"
```

drpatelh:
Just means you wont have to evaluate `${params.custom_config_version}` too.

apeltzer:
That should be good yes - however all older pipelines having the current version implemented had these `${params.custom_config_version}` set. The new approach will define `custom_config_base`, so should work fine, too 👍

sven1103:
sorry, this was parallel work to what @drpatelh suggested :) Guess both does the job.

PR #24, opened 2019-03-06
Include config integration test

sven1103:


sven1103:
nasty API rate limits -.-

ewels:
Looks good! What do you think about using a minimal pipeline instead of running an actual nf-core pipeline? Would be quicker and we could get it to print out some variable saved in the test profile. If we test that this is printed we then ensure that the full config is being loaded properly..

Could also be nice to test each profile somehow, to make sure that they’re all being parsed without errors?

Could also check that the descriptive params are being set in each profile too.

sven1103:
> What do you think about using a minimal pipeline instead of running an actual nf-core pipeline?
 
uh yes, this is a nice idea! 

> Could also be nice to test each profile somehow, to make sure that they’re all being parsed without errors?

Is there a nice way of loading every profile in a hardcoded fashion? I mean, I can parse the config with the profiles, but it feels dirty.

apeltzer:
> Is there a nice way of loading every profile in a hardcoded fashion? I mean, I can parse the config with the profiles, but it feels dirty.

I think we'd need to load the `-profile xyz` and then check some of the `params.config_profile_description` , `params.config_profile_contact`, `params.config_profile_url` that are set by all of the profiles available here in the `configs` repository

ewels:
Yes, but I'm not sure that there's an easy way to know the name of the profiles to be able to call `-profile xyz` - I guess this is what @sven1103 meant (I have been ruminating on the same thing).

I wondered about calling `nextflow config` but I don't think that this gives you the available profile names. So my best guess is just doing some bash grepping to pull out the profile names from the file first. That or hardcoding their names.

Phil

apeltzer:
Yeah, that is also what I meant! I think it's not possible and just hardcoding available profiles in there is one approach (or grepping for all available ones using a simple script, which I like more since it scales to new profiles automatically too)...

sven1103:
@apeltzer how long did you google and try and error 😂

apeltzer:
Ah wasn't that bad https://regex101.com/ is your friend (thanks @ewels )

sven1103:
I have no idea why it exits with 123, locally I also get an exit code of 1, but I don't understand why.

sven1103:
ha, found it. @apeltzer the expression is wrong, because it creates `uppmax` two times, but it should be `uppmax` and `uppmax_devel`. That's why `xargs` returns with 123. 🤦‍♂️ nasty issue to track down

sven1103:
Hm, still exits with 123, although profiles are correct now. Any ideas? @ewels @apeltzer

sven1103:
🙌 it worked now

sven1103:
`grep "{.*includeConfig.*[a-z]*\.config\"" ${TRAVIS_BUILD_DIR}/nfcore_custom.config | tr -s ' ' | cut -d " " -f 2 | xargs -I {} nextflow run ${TRAVIS_BUILD_DIR}/configtest.nf -profile {}
`

does the job know

ewels:
Super nice! 🚀

sven1103:
thank you :D

apeltzer:
Tests are running through fine :-)

PR #26, opened 2019-03-07
Rolling out iGenomes now

apeltzer:
Since we now synced it - we can roll it out 👍


apeltzer:
Since its only affecting us (for sure - its our cluster...) i'll merge after our admin had a look at this ;-)

ewels:
Nice! 🎉

PR #29, opened 2019-03-11
iGenomes Rollout on "Binac" too

apeltzer:
Yup, we're ready!!

### DIFF
--- a/conf/binac.config
+++ b/conf/binac.config
@@ -17,7 +17,8 @@ process {
 }
 
 params {
-  igenomesIgnore = true
+  igenomesIgnore = false
+  igenomes_base = '/beegfs/work/igenomes'
   max_memory = 128.GB
   max_cpus = 28
   max_time = 48.h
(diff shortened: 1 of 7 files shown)


ewels:
Tests failing because merging to `master`. Switched to `dev`.

Not sure that we need this test if we're not doing releases and we always want the latest (stable) code on `master`... Tests should run on PRs to make sure that we're not merging things that break. What do you think about removing that part of the travis config @sven1103?

ewels:
@apeltzer - I don't think that the tests will properly rerun on the switched branch. Can you maybe pop a new commit on the end to make Travis run again please?

ewels:
mini adventure

conf/binac.config
```diff
@@ -17,7 +16,8 @@ process {
 }
 
 params {
-  igenomesIgnore = true
+  igenomesIgnore = false
```
ewels: I think this is the default so can be removed in the future if you want...

apeltzer: True, can drop it as well :-)

PR #32, opened 2019-03-12
MENDEL: remove singularity support

phue:
This is a follow-up to the discussion we had in https://github.com/nf-core/configs/pull/16#discussion_r263058927

I decided to remove singularity entirely from the documentation for the following reasons:

* the kernel running on the compute nodes does not support overlayfs und thus containers downloaded from
  public registries won't run because they lack the /lustre mountpoint

* building containers manually using the Singularity file would work, however nf-core pipelines will stop
  providing these files in future releases

### DIFF
--- a/conf/mendel.config
+++ b/conf/mendel.config
@@ -11,7 +11,7 @@ manifest {
 }
 
 process {
-  beforeScript = {'module load Singularity; module load Miniconda3'}
+  beforeScript = 'module load Miniconda3/4.6.7'
   executor = 'pbspro'
   clusterOptions = { "-P $params.project" }
 }
--- a/docs/mendel.md
+++ b/docs/mendel.md
@@ -2,27 +2,15 @@
 
 All nf-core pipelines have been successfully configured for use on the MENDEL CLUSTER at the Gregor Mendel Institute (GMI).
 
-To use, run the pipeline with `-profile conda,mendel`. This will download and launch the [`mendel.config`](../conf/mendel.config) which has been pre-configured with a setup suitable for the MENDEL cluster. A Conda environment will be created automatically and software dependencies will be downloaded from ['bioconda'](https://bioconda.github.io/).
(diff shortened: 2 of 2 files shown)


phue:
Hi @ewels ,

Thanks alot for your input! I've been playing around with this some more but it really doesn't seem to work.
As much as I would like to have singularity as an option, I think conda is fine for now.

From your post in the other thread:

> ```shell
> singularity pull --name nf-core-rnaseq-1.1.simg docker://nfcore/rnaseq:1.1
> sudo singularity exec --writable nf-core-rnaseq-1.1.simg mkdir /lustre
> ```
> 
> See http://singularity.lbl.gov/archive/docs/v2-3/docs-changing-containers

The `sudo` is the problem here. It would need to be done manually on another machine where root access is available.


> Or my child container suggestion was to make your own singularity build script for each pipeline which would look like this:
> 
> ```
> Bootstrap: docker
> From: nfcore/rnaseq:1.1
> 
> %post
>     mkdir /lustre
> ```

That's an interesting option but where should these recipes live? In the https://github.com/nf-core/configs repo?

Honestly I think that it's not worth the effort. Building the container needs to be done on another machine once again and is a documentation nightmare.
It seems to me that currently only our infrastructure has the problem and the current cluster will be replaced by mid of the year. Hopefully the new one will properly support singularity.

> I have a vague memory of @remiolsen having some kind of nicer run-time nextflow config option for binding directories in Singularity for non-overlayfs people too. But I can't remember the details, or whether it worked. Basically in your nextflow config file having this I think:
> 
> ```nextflow
> singularity.engineOptions = '-B /lustre:/lustre'
> ```

Also doesn't work because it would try to bind a non-existent path again.

I found a hacky way to add the mountpoint to the container which doesn't require root privileges. But it's not a nice solution and I would have to convince nextflow to execute this right after the container was pulled, in order to do this automatically for all pipelines. :

```shell
singularity pull --name nf-core-rnaseq-1.1.simg docker://nfcore/rnaseq:1.1
mkdir ./lustre; tar -cp ./lustre | singularity image.import nf-core-rnaseq-1.1.simg
```

ewels:
Sounds good - thanks for trying this stuff out, it's good to know for future users with the same problem! Replacing the cluster is a slightly extreme way of solving it, but probably the best approach 😅

PR #33, opened 2019-03-13
Removed params.config_profile_name, added params.hostnames

ewels:
`params.config_profile_name` didn't seem to add anything beyond what we already had in the profile name, so I got rid of it.

I also added a new groovy map on the main config file called `params.hostnames`... Why? All will soon become clear...

### DIFF
--- a/conf/crick.config
+++ b/conf/crick.config
@@ -1,8 +1,7 @@
 //Profile config names for nf-core/configs
 params {
-  config_profile_name = 'CRICK'
   config_profile_description = 'The Francis Crick Institute CAMP HPC cluster profile provided by nf-core/configs.'
-  config_profile_contact = 'Harshil Patel (@drpatelh )'
+  config_profile_contact = 'Harshil Patel (@drpatelh)'
   config_profile_url = 'https://www.crick.ac.uk/research/platforms-and-facilities/scientific-computing/technologies'
 }
 
--- a/nfcore_custom.config
+++ b/nfcore_custom.config
@@ -23,7 +23,7 @@ profiles {
   phoenix      { includeConfig "${params.custom_config_base}/conf/phoenix.config" }
   shh          { includeConfig "${params.custom_config_base}/conf/shh.config" }
   uct_hex      { includeConfig "${params.custom_config_base}/conf/uct_hex.config" }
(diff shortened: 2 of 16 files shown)


conf/crick.config
```diff
@@ -1,8 +1,7 @@
 //Profile config names for nf-core/configs
 params {
-  config_profile_name = 'CRICK'
   config_profile_description = 'The Francis Crick Institute CAMP HPC cluster profile provided by nf-core/configs.'
-  config_profile_contact = 'Harshil Patel (@drpatelh )'
```
drpatelh: Ive been meaning to remove that bloody space for ages :+1: :laughing:

ewels:
x-ref: https://github.com/nf-core/tools/pull/278

nfcore_custom.config
```diff
@@ -23,9 +23,17 @@ profiles {
...
   uppmax       { includeConfig "${params.custom_config_base}/conf/uppmax.config" }
   uzh          { includeConfig "${params.custom_config_base}/conf/uzh.config" }
 }
+
+// If user hostnames contain one of these substring and they are
+// not running the associated profile, it will trigger a warning message
+// Should be defined here for all profiles (not within profile config)
+params {
+  // This is a groovy map, not a nextflow parameter set
+  hostnames = [
```
drpatelh: Think Ive figured out whats going on by looking at the PR in `nf-core/tools`.

We would all have to explicitly specify a `hostname` value here right e.g.  
```nextflow
params {
￼	  // This is a groovy map, not a nextflow parameter set
￼	  hostnames = [
              uppmax: ['.uppmax.uu.se'],
              crick: ['.thecrick.org']
￼	  ]
￼	}


```

drpatelh: Everything else looks good to me.

ewels: Yes - that is correct :) So more PRs needed to add all of those, but seems to work nicely from my initial testing. Also needs PRs to every pipeline to add the function that uses this. Now it's being added to the template it should slowly propagate out though..

ewels: This is what I get with test data on the dev rnaseq pipeline now:
image

PR #36, opened 2019-03-15
add check_max function 

KochTobi:
Add `check_max(obj,type)` function to [`nextflow.config`](https://github.com/nf-core/configs/blob/dev/nextflow.config) file

copied from [nf-core/rnaseq](https://github.com/nf-core/rnaseq/blob/dev/nextflow.config)

Is this the right place for that? Please feel free to edit this PR and leave a comment with your thoughts.

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -1,1 +1,34 @@
 includeConfig("nfcore_custom.config")
+
+// Function to ensure that resource requirements don't go beyond
+// a maximum limit
+def check_max(obj, type) {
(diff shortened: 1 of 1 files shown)


KochTobi:
Since this should be included in every nf-core pipeline I guess I can close this PR?

ewels:
Yes, and I don't think that you can include functions in imported configs like this (I tried with the new hostnames thing and it didn't work sadly).

PR #38, opened 2019-03-18
add queueSize limited to 5 for SHH short queue

maxibor:
Limit queueSize to 5 simultaneous jobs.
Equivalent to a SLURM Job array of 5 simultaneous jobs.
@jfy133

### DIFF
--- a/conf/shh.config
+++ b/conf/shh.config
@@ -13,6 +13,7 @@ singularity {
 process {
     executor = 'slurm'
     queue = 'short'
+    queueSize = 5
 }
 
 params {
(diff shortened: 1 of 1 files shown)


jfy133:
I'll think about it tonight. I think 5 might be too restrictive, but we can look into largest default EAGER process requirements and adapt accordingly.

jfy133:
After looking at the EAGER2 base.config for default settins. Could you switch to 16 instead? That would fill a single CDAG node of memory and that is reasonable IMO.

maxibor:
> memory

Done

PR #43, opened 2019-05-04
BINAC SMP / Short / Long Updates

apeltzer:
Hi @andreas-wilm - would you please have a look and amend my changes? 

That way all nf-core pipelines will be useable on the Aquila cluster 👍


ewels:
I suspect that we need a new person responsible for this config.

apeltzer:
Yeah I just stripped this out of the nf-core/bacass pipeline and didn't want to completely just delete it without the option to make it work 👍

apeltzer:
@drpatelh Can you have a look at this? I don't have anyone to have a look at the Aquila config but my Commits just got merged into here too - so this should not get lost. if we don't have anybody from Singapore taking care of this cluster config, we can also delete it though ...

drpatelh:
If we are being really picky :sweat_smile: 
`BIGPURPLE` should be before `BINAC`
`UPPMAX should be before `UPPMAX-DEVEL`

drpatelh:
Can also change this line in `pasteur.config`:
https://github.com/nf-core/configs/pull/58#pullrequestreview-268454086

MaxUlysse:
To be honest, whenever I need to sort, I use the `sort` in my editor, very useful for big lists and such issues.

drpatelh:
Not sure there is any point in adding it if no-one is using it now.

apeltzer:
Done and Done and Done :-)

apeltzer:
Some minor cleanups on docs in pasteur configs.

Addition of new queue logic for `binac`.

apeltzer:
Fixes #75

PR #45, opened 2019-05-31
[WIP] Add CZ Biohub

olgabot:
Haven't tested that it works yet


apeltzer:
Two configs are the way to go then 👍

apeltzer:
Sorry, wasn't on a notebook for days (public holiday over here): You could add two configs then, one `czbiohub` and another one `czbiohub-lsf` or similar if you want to,

PR #53, opened 2019-06-28
Multiple uppmax clusters

alneberg:
There are several clusters at [Uppmax](https://www.uppmax.uu.se/resources/systems/) with slightly different hardware. This PR adds these as separate profiles which all share a common `uppmax.conf` base config. The general uppmax profile is kept since a lot of documentation currently refers to this one. 

The memory parameter is transformed with the 1000/1024 factor to avoid requesting more MB memory than what is available.

### DIFF
--- a/conf/uppmax-bianca.config
+++ b/conf/uppmax-bianca.config
@@ -1,0 +1,4 @@
+params {
+  max_memory = 125.GB
+  max_cpus = 16
+}
--- a/conf/uppmax-irma.config
+++ b/conf/uppmax-irma.config
@@ -1,0 +1,4 @@
+params {
+  max_memory = 250.GB
+  max_cpus = 16
+}
--- a/conf/uppmax-rackham.config
+++ b/conf/uppmax-rackham.config
@@ -1,0 +1,4 @@
+params {
+  max_memory = 125.GB
+  max_cpus = 20
+}
--- a/docs/uppmax.md
+++ b/docs/uppmax.md
@@ -3,13 +3,13 @@
 All nf-core pipelines have been successfully configured for use on the Swedish UPPMAX clusters.
 
 ## Using the UPPMAX config profile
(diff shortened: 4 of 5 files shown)


ewels:
I thought about this before, but decided to leave it - Rackham and Bianca are the same, and we use our own [custom configs](https://github.com/NationalGenomicsInfrastructure/irma-provision/blob/master/roles/rnaseq/templates/rnaseq_site.config) for NGI deployments on Irma (and no-one else uses that cluster). So I'm not sure if this is really needed..?

alneberg:
@ewels, sorry I missed your comment and forgot about this PR 🤦‍♂ . Ok, I agree adding irma is not really that useful. But rackham and bianca are actually slightly different, with 16 max_cpus on bianca and 20 max_cpus on rackham. What do you think?

MaxUlysse:
also bianca being offline, it would be interesting to have more detailed docs on how to use the offline configs.
I had users that were slightly confused on how to do it.

ewels:
Yes, docs could do with a bit more on that. I've never really used Bianca so they're a bit thin. PRs welcome! 😉 https://github.com/nf-core/configs/blob/master/docs/uppmax.md#running-offline-with-bianca

@alneberg - we could be potentially clever about this and dynamically set `params.max_cpus` based on the hostname 🚀 What do you think? I'd prefer not to proliferate the configs if possible, I like the simplicity of having just `-profile uppmax`...

MaxUlysse:
Where are you on that one @alneberg?

alneberg:
I like the suggestion by @ewels. It's quite different from this PR though, so it should maybe start from the beginning?

MaxUlysse:
As you prefer, it's your PR

MaxUlysse:
I'm pretty sure we can close this one, now that #100 has been merged

PR #55, opened 2019-07-05
Updating CCGA config file

marchoeppner:
Updating some details of the ccga config file, switching to use singularity.


marchoeppner:
Sorry, missed the markup and updated it.

PR #57, opened 2019-07-29
Adding CCGA DX cluster config

marchoeppner:
We have a second, smaller cluster on which we also want to run some of the nf-core pipelines. Config file included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Currently documentation is available for the following clusters:
 * [BINAC](docs/binac.md)
 * [BIGPURPLE](docs/bigpurple.md)
 * [CCGA](docs/ccga.md)
-* [CCGA-DX]/docs/ccga_dx.md)
+* [CCGA_DX](/docs/ccga_dx.md)
 * [CFC](docs/binac.md)
 * [CRICK](docs/crick.md)
 * [GIS](docs/gis.md)
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ Currently documentation is available for the following clusters:
 * [BINAC](docs/binac.md)
 * [BIGPURPLE](docs/bigpurple.md)
 * [CCGA](docs/ccga.md)
+* [CCGA-DX]/docs/ccga_dx.md)
 * [CFC](docs/binac.md)
 * [CRICK](docs/crick.md)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -85,6 +85,7 @@ Currently documentation is available for the following clusters:
 * [BINAC](docs/binac.md)
 * [BIGPURPLE](docs/bigpurple.md)
 * [CCGA](docs/ccga.md)
+* [CCGA-DX]/docs/ccga_dx.md)
```
drpatelh:
```suggestion
* [CCGA_DX](/docs/ccga_dx.md)
```

docs/ccga_dx.md
```diff
@@ -0,0 +1,9 @@
+# nf-core/configs: CCGA DX Configuration
+
+Deployment and testing of nf-core pipelines at the CCGA DX cluster is on-going. 
+
+To use, run the pipeline with `-profile ccga_dx`. This will download and launch the [`ccga.config`](../conf/ccga.config) which has been pre-configured with a setup suitable for the CCGA cluster. Using this profile, a docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.
```
drpatelh:
```suggestion
To use, run the pipeline with `-profile ccga_dx`. This will download and launch the [`ccga_dx.config`](../conf/ccga_dx.config) which has been pre-configured with a setup suitable for the CCGA cluster. Using this profile, a docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.
```

drpatelh: Do you think its also worth adding some description regarding the iGenomes source you have available locally?
https://github.com/nf-core/configs/blob/master/docs/crick.md

drpatelh: Could also add the same to `ccga.md`

marchoeppner: In principle, this is a fair point, in practice this particular cluster is my "personal playground". I can maybe add it to the ccga.config at some point, which is used by several people.

drpatelh: Ok. Fair point. Would be good if you add it in when you know where its going to reside 👍

PR #63, opened 2019-08-02
Dummy Pull Request to test fixes for czbiohub config testing

tobsecret:
Trying out the possibility of using a TEST environmental variable that makes it so it passes travis tests while it would still be usable by the czbiohub folks.

`travis.yml`
```
env:
  - NXF_VER='18.10.1' SCRATCH='~' TEST=true
  - NXF_VER='' SCRATCH='~' TEST=true
```
`czbiohub_aws.config`
```
  executor = TEST ? 'local' : "awsbatch"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,9 @@ install:
   - mkdir -p ${TRAVIS_BUILD_DIR}/tests && cd ${TRAVIS_BUILD_DIR}/tests
 
 env:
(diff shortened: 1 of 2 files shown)


ewels:
The problem is that the tests then fail for _everyone_, so it's difficult to spot problems with other people's configs..

jfy133:
Hi @tobsecret  I'm closing this as stale as it's not been updated in over a year. Feel free to re-open as necessary.

PR #66, opened 2019-08-07
Updating CCGA configs - adding cache dir and mount points

marchoeppner:
Noticed a singularity mounting issue on ccga_dx; and adding a cache dir for the main production cluster ccga.

### DIFF
--- a/conf/ccga_dx.config
+++ b/conf/ccga_dx.config
@@ -13,7 +13,7 @@ params {
 
 singularity {
   enabled = true
-  runOoptions = "-B /mnt"
+  runOptions = "-B /mnt"
 }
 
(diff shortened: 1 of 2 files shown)


conf/ccga_dx.config
```diff
@@ -13,6 +13,7 @@ params {
 
 singularity {
   enabled = true
+  runOoptions = "-B /mnt"
```
apeltzer: This looks like a typo! 
```suggestion
  runOptions = "-B /mnt"
```

marchoeppner: Ah...oh well, should be fixed in most recent commit.

PR #67, opened 2019-09-22
add cbe config

phue:
This adds a config for the new cluster at the Vienna BioCenter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,4 @@ script:
       tr -s ' ' | \
       cut -d " " -f 2 | \
       grep -v "czbiohub_aws" | \
-      xargs -I {} nextflow run ${TRAVIS_BUILD_DIR}/configtest.nf -profile {}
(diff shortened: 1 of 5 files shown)


phue:
Looks like tests fail because it tries to download the new config from the master branch (which doesn't contain the file yet).

 I guess we could just add a test.config that points `params.custom_config_base` to the local clone of the repo on the travis build machine?

drpatelh:
Yeah you're right @phue :+1:  I saw this when I was fixing the broken Travis CI tests for this repo. @sven1103 implemented these tests initially. Any other suggestions where we can test the config before it's actually added to the repo?

phue:
I added a commit that sets `params.custom_config_base` to ${TRAVIS_BUILD_DIR} when calling nextflow.

The alternative would be to also make use of a `dev` branch for this repo. For example push to `dev` first and then only set `params.custom_config_version='dev'` when testing PRs against `master`

apeltzer:
This is however also working perfectly fine when you merge this to `master` as the `TRAVIS_BUILD_DIR` will then be `master`, too. Thanks for adding this !

PR #68, opened 2019-10-08
Add Singularity Cache Directory

apeltzer:


conf/cfc.config
```diff
@@ -7,6 +7,7 @@ params {
 
 singularity {
   enabled = true
+  cacheDir = '/nfsmounts/containers'
```
apeltzer:
```suggestion
  cacheDir = '/nfsmounts/container'
```

drpatelh:
@apeltzer Something we discussed a while back and I havent got around to testing because I dont know if I am allowed to use the file paths on our cluster (for security reasons).

I thought it may be nice to use something like:
`cacheDir = "/path/to/local/dir/${workflow.manifest.name}/v${workflow.manifest.version}/"`

Where you would get separate folders for containers from different pipelines, and so they would be organised better. Putting all of the containers in a single directory may get messy especially when we move to modules so probably something we need to think about now?

If so, fancy testing it?

apeltzer:
Wow @drpatelh ! Sure I fancy testing this 🎉

drpatelh:
Thanks :100:

PR #71, opened 2019-10-10
Add profiles that should be shared with all pipelines to configs

MaxUlysse:
I do think `test` and `debug` should be the only profiles to be specified at the pipeline level.

So I'd like to include the following profile here:
- awsbatch

I still think `conda`, `docker` and `singularity` should be included there, but I agree that we can do that one first, and then we'll see.

### DIFF
--- a/conf/awsbatch.config
+++ b/conf/awsbatch.config
@@ -1,0 +1,14 @@
+//Nextflow config file for running on AWS batch
+
+params {
+  config_profile_name = 'AWSBATCH'
+  config_profile_description = 'AWSBATCH Cloud Profile'
+  config_profile_contact = 'Alexander Peltzer (@apeltzer)'
+  config_profile_url = 'https://aws.amazon.com/batch/'
+}
+
+aws.region = params.awsregion
+process.executor = 'awsbatch'
+process.queue = params.awsqueue
+executor.awscli = '/home/ec2-user/miniconda/bin/aws'
+params.tracedir = './'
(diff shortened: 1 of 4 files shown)


conf/awsbatch.config
```diff
@@ -0,0 +1,14 @@
+//Nextflow config file for running on AWS batch
+
+params {
+  config_profile_name = 'AWSBATCH'
+  config_profile_description = 'AWSBATCH Cloud Profile'
+  config_profile_contact = 'Alexander Peltzer (@apeltzer)'
+  config_profile_url = 'https://aws.amazon.com/batch/'
+}
+
+aws.region = params.awsregion
```
ewels: This line causes the tests to fail 😱 

https://travis-ci.com/nf-core/configs/jobs/258237834#L362

conf/awsbatch.config
```diff
@@ -0,0 +1,14 @@
...
+process.executor = 'awsbatch'
+process.queue = params.awsqueue
```
ewels: This line will also cause problems.

conf/awsbatch.config
```diff
@@ -0,0 +1,14 @@
...
+executor.awscli = '/home/ec2-user/miniconda/bin/aws'
+params.tracedir = './'
```
ewels: Any reason not to group this together with the other `params` at the top?

MaxUlysse: None, I just copied the config file we had on sarek

PR #77, opened 2019-10-28
deprecate igenomesIgnore in favor of igenomes_ignore

phue:
as discussed in https://github.com/nf-core/methylseq/pull/101

Let's keep both for now until all the pipelines have been synced.


drpatelh:
Amazing! Thanks @phue

PR #79, opened 2019-10-28
Fix for BinAC profile / resources etc pp

apeltzer:


apeltzer:
Now finally tested and successfully adjusted on `Binac`.

PR #82, opened 2019-11-06
documentation on how to fix bianca "Requested node configuration is not available" problem

arontommi:
Bianca only allows 7 GB per core, so max_mem of 128 GB results in an "Requested node configuration is not available" error.

### DIFF
--- a/docs/uppmax.md
+++ b/docs/uppmax.md
@@ -26,7 +26,7 @@ If running on Bianca, you will have no internet connection and these configs wil
 Please use the nf-core helper tool on a different system to download the required pipeline files, and transfer them to bianca.
 This helper tool bundles the config files in this repo together with the pipeline files, so the profile will still be available.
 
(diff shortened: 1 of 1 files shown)


MaxUlysse:
Hi @arontommi good spot.
Can't we get a fat node on bianca?
And just out of curiosity, which pipeline where you running?

arontommi:
Thanks,
Yes you could request a fat node instead
I was running sarek (vc with strelka in this case )

ewels:
There's already docs there about using fat nodes, so I'm happy with this. Thanks!

ewels:
Tests are failing due to the missing `MENDEL` config, which was removed in #80. Not sure why though, I've retriggered them - maybe they will pass on the second attempt 🤔

ewels:
Switched PR to be against the master branch.

ewels:
Tests fixed 🎉

PR #89, opened 2019-11-26
Fix issues with awsbatch profile

MaxUlysse:
Should fix tests broken by #71

### DIFF
--- a/conf/awsbatch.config
+++ b/conf/awsbatch.config
@@ -1,14 +1,16 @@
 //Nextflow config file for running on AWS batch
 
 params {
-  config_profile_name = 'AWSBATCH'
-  config_profile_description = 'AWSBATCH Cloud Profile'
+  awsqueue = ''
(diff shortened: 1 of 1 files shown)


conf/awsbatch.config
```diff
@@ -1,14 +1,16 @@
 //Nextflow config file for running on AWS batch
 
 params {
-  config_profile_name = 'AWSBATCH'
-  config_profile_description = 'AWSBATCH Cloud Profile'
+  awsqueue = ''
```
MaxUlysse:
```suggestion
  awsqueue = false
```

conf/awsbatch.config
```diff
@@ -1,14 +1,16 @@
...
+  awsregion = ''
```
MaxUlysse:
```suggestion
  awsregion = 'eu-west-1'
```

PR #93, opened 2019-11-26
Updated max resources for SHH SDAG 

jfy133:
Note: currently this exceeds limits for CDAG, but is required for SDAG (e.g. if running MALT in EAGER).

Suggestions on if there is a way to specify different clusters? Should we make two configs - shh-sdag and shh-cdag, for example (@apeltzer @maxibor )?


apeltzer:
I'd say have two configs for two systems is fine and good practice.

apeltzer:
Markdownlinting fails @jfy133

jfy133:
> I'd say have two configs for two systems is fine and good practice.

Will seperate this evening :+1:

PR #97, opened 2019-11-27
Mini adjustment for `highmem` queue on denbi

apeltzer:
Since we have two queues now on de.NBI (512GB nodes on highmem, 64GB ones on batch), we need to make this change to get the big nodes used :-)


apeltzer:
@skrakau please review!

PR #98, opened 2019-11-27
Set awsqueue parameter properly

olgabot:


apeltzer:
Still need to find a proper test for the czi_awsbatch profiles as these are simply ignored atm.

PR #100, opened 2019-12-05
Improve UPPMAX configs

MaxUlysse:
EDIT:
- replace `-profile uppmax_devel` by `-profile uppmax,devel`
- automatic assignation of resources based on hostnames
- update docs

Let's simplify some configs.
So let's start by getting ridd of `uppmax_devel`:

```Bash
≻nextflow run . -profile uppmax
[...]
Config Profile    : uppmax
Config Description: Swedish UPPMAX cluster profile provided by nf-core/configs.
Config Contact    : Phil Ewels (@ewels)
Config URL        : https://www.uppmax.uu.se/
```

```Bash
≻nextflow run . -profile uppmax,devel
[...]
Profile    : uppmax,devel
Config Description: Testing & development profile for UPPMAX, provided by nf-core/configs.
Config Contact    : Phil Ewels (@ewels)
Config URL        : https://www.uppmax.uu.se/
```

PR #101, opened 2019-12-05
Condensing SHH profile(s) with sub-profiles

jfy133:
Following the (surprisingly simple) wizardary of https://github.com/nf-core/configs/pull/100


conf/shh.config
```diff
@@ -22,9 +22,25 @@ executor {
...
   max_time = 720.h
   //Illumina iGenomes reference file path
   igenomes_base = "/projects1/public_data/igenomes/"
 }
+
+
+profiles {
+  cdag {
+      config_profile_description = 'MPI-SHH CDAG profile, provided by nf-core/configs.'
+      max_memory = 256.GB
```
MaxUlysse:
```suggestion
```

MaxUlysse: already there by default

conf/shh.config
```diff
@@ -22,9 +22,25 @@ executor {
...
+      max_cpus = 32
```
MaxUlysse:
```suggestion
```

conf/shh.config
```diff
@@ -22,9 +22,25 @@ executor {
...
+      max_time = 720.h
```
MaxUlysse:
```suggestion
```

conf/shh.config
```diff
@@ -22,9 +22,25 @@ executor {
...
+  }
+  sdag {
+      config_profile_description = 'MPI-SHH SDAG profile, provided by nf-core/configs.'
+      max_memory = 2.TB
+      max_cpus = 128
+      max_time = 720.h
```
MaxUlysse:
```suggestion
```

jfy133:
FYI: I would prefer to keep the `cdag` sub-profile for consistency (and we might have different specs relatively soon), even if it follows the 'default' SHH parameters.

PR #117, opened 2020-02-07
Cleanup config

d4straub:


d4straub:
Would like to have Gisela's comment here as well since this could potentially crush all our pipelines on BinAC

ggabernet:
LGTM too, thanks :)

PR #121, opened 2020-02-20
Uppmax

MaxUlysse:
- add UPPMAX specific sarek config

@aanil we can add any params in `conf/pipeline/sarek/uppmax.config`
For `irma` specific params or configs, it can go inside the ` if (hostname ==~ "i.*") {}` block


README.md
```diff
@@ -196,7 +196,7 @@ We will be notified automatically when you have created your pull request, and p
 [Fork](https://help.github.com/articles/fork-a-repo/) the [`nf-core/configs`](https://github.com/nf-core/configs/) repository to your own GitHub account.
 And add or edit the following files in the local clone of your fork.
 
-* `pipeline/<PIPELINE>.config`
+- `pipeline/<PIPELINE>.config`
```
MaxUlysse:
```suggestion
* `pipeline/<PIPELINE>.config`
```

conf/pipeline/sarek/uppmax.config
```diff
@@ -0,0 +1,17 @@
+// sarek/uppmax specific profile config
+
+params {
+  config_profile_contact = 'Maxime Garcia (@MaxUlysse)'
+  config_profile_description = 'nf-core/sarek uppmax profile provided by nf-core/configs'
+}
+
+if (hostname ==~ "i.*") {
+  singularity {
+      cacheDir = "${params.containers_path}"
```
ewels:
```suggestion
      cacheDir = params.containers_path
```
I think you maybe need to set a default value for this as well, to avoid warnings on other systems? Not sure, maybe not necessary.

MaxUlysse: That should only happen on irma, so we should be fine with that.

conf/pipeline/sarek/uppmax.config
```diff
@@ -0,0 +1,17 @@
+// sarek/uppmax specific profile config
+
+params {
+  config_profile_contact = 'Maxime Garcia (@MaxUlysse)'
+  config_profile_description = 'nf-core/sarek uppmax profile provided by nf-core/configs'
+}
+
+if (hostname ==~ "i.*") {
+  singularity {
+      cacheDir = params.containers_path
```
alneberg: I think we said that Irma-specific configs we might as well put in the irma-provision repositories. The users for that cluster is limited to NGI anyway.

conf/pipeline/sarek/uppmax.config
```diff
@@ -0,0 +1,17 @@
...
+  }
+}
+
+params {
+  igenomeIgnore = true
+  genomes_base = params.genome == 'GRCh37' ? '/sw/data/uppnex/ToolBox/ReferenceAssemblies/hg38make/bundle/2.8/b37' : '/sw/data/uppnex/ToolBox/hg38bundle'
```
alneberg: Don't get why we have two different locations but fair enough.

conf/pipeline/sarek/uppmax.config
MaxUlysse: I think the current idea is to get everything out of irma provision

conf/pipeline/sarek/uppmax.config
MaxUlysse: That's only while iGenomes is not updated

PR #128, opened 2020-03-06
Update cfc.config

mseybold:
We just added some new HighMem nodes to the cfc, Partition/queue name is qbic, each node has 128Cores and 2TB of RAM.
The old queue (Name compute) is set to default by slurm.
Each process over 60GB of RAM or over 20cpus should run on those new nodes.
So added the queue State and adjusted the max_memory and max_cpu setting.


mseybold:
The link in the markdown doc leads to binac for cfc :)

PR #140, opened 2020-03-11
Update Munin specific configs

MaxUlysse:
- Update Sarek/Munin specific configs
- Add rnafusion/Munin specific configs
- Fix links to config files in pipeline/specific docs
- Markdown lint

### DIFF
--- a/conf/pipeline/sarek/munin.config
+++ b/conf/pipeline/sarek/munin.config
@@ -1,4 +1,4 @@
- // sarek/munin specific profile config
+// sarek/munin specific profile config
 
 params {
   // Specific nf-core/configs params
--- a/conf/pipeline/sarek/munin.config
+++ b/conf/pipeline/sarek/munin.config
@@ -1,4 +1,4 @@
-// Profile config names for nf-core/configs
+ // sarek/munin specific profile config
 
 params {
   // Specific nf-core/configs params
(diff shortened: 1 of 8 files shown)


MaxUlysse:
@nf-core/core Not actually sure why, but I add markdownlint issues with the unordered list on my computer, so that's why I changed the `*` by `-`

apeltzer:
> 
> 
> @nf-core/core Not actually sure why, but I add markdownlint issues with the unordered list on my computer, so that's why I changed the `*` by `-`

I had that too - I think the default by markdownlint is to request the `-` though I like the `*` more. I think you can configure your local markdownlint too to tkae that as new default ...?

MaxUlysse:
> > @nf-core/core Not actually sure why, but I add markdownlint issues with the unordered list on my computer, so that's why I changed the `*` by `-`
> 
> I had that too - I think the default by markdownlint is to request the `-` though I like the `*` more. I think you can configure your local markdownlint too to tkae that as new default ...?

I think I have the `-` in sarek now, so I'm kinda used to that one, but actually I don't really care.
I'll revert the `*` to `-` changes I made, and I'll ask for a decision on that particular Markdown formatting matter somewhere else

apeltzer:
For reference, this is the rule that causes the trouble there: 

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md004---unordered-list-style

apeltzer:
Sorry :-( 

I think this is the rule we'd need to adjust in the `markdownlint.yml`:

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md004---unordered-list-style

conf/pipeline/sarek/munin.config
```diff
@@ -1,22 +1,28 @@
-// Profile config names for nf-core/configs
+ // sarek/munin specific profile config
```
MaxUlysse:
```suggestion
// sarek/munin specific profile config
```

MaxUlysse:
Actually tests are working fine, so no need to change anything...

PR #143, opened 2020-03-23
add cfc_dev profile

ggabernet:
- Adding cfc_dev profile to be able to run pipelines with -r dev without caching containers
- Adding warnings when people are running pipelines in binac or cfc without the corresponding profiles

### DIFF
--- a/nfcore_custom.config
+++ b/nfcore_custom.config
@@ -17,6 +17,7 @@ profiles {
   ccga_dx      { includeConfig "${params.custom_config_base}/conf/ccga_dx.config" }
   ccga_med     { includeConfig "${params.custom_config_base}/conf/ccga_med.config" }
   cfc          { includeConfig "${params.custom_config_base}/conf/cfc.config" }
+  cfc_dev      { includeConfig "${params.custom_config_base}/conf/cfc_dev.config" }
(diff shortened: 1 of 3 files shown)


nfcore_custom.config
```diff
@@ -47,7 +48,9 @@ profiles {
 params {
   // This is a groovy map, not a nextflow parameter set
   hostnames = [
+    binac: ['.binac.uni-tuebingen.de'],
```
apeltzer: Are these valid now? They have not been when I checked last time? Same below for CFC ...?

ggabernet: That's contained in the Hostnames when I log in

ggabernet: doesn't it work like this?

apeltzer: It does :-) Then they added that apparently :-)

apeltzer: Maybe check if nodes have that too 😓

ewels: I don't think that it matters about the nodes here.. As I remember, this `hostnames` array is only used to warn the user if they are not using the relevant config profile when they launch Nextflow (and maybe again if it exits with an error).

ggabernet: ok, then it should be fine, thanks, I'll edit it again if it does not work :)

PR #145, opened 2020-03-31
Adding UPPMAX specific config for ampliseq

erikrikarddaniel:
I added what I hope is a correct config to use ampliseq with fat nodes on UPPMAX.

### DIFF
--- a/pipeline/ampliseq.config
+++ b/pipeline/ampliseq.config
@@ -11,3 +11,7 @@
 profiles {
   binac { includeConfig "${params.custom_config_base}/conf/pipeline/ampliseq/binac.config" }
 }
+
+profiles {
+  uppmax { includeConfig "${params.custom_config_base}/conf/pipeline/ampliseq/uppmax.config" }
+}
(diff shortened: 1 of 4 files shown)


pipeline/ampliseq.config
```diff
@@ -10,4 +10,8 @@
 
 profiles {
   binac { includeConfig "${params.custom_config_base}/conf/pipeline/ampliseq/binac.config" }
-}
\ No newline at end of file
+}
```
MaxUlysse:
```suggestion
```

pipeline/ampliseq.config
```diff
@@ -10,4 +10,8 @@
...
+
```
MaxUlysse:
```suggestion
```

pipeline/ampliseq.config
```diff
@@ -10,4 +10,8 @@
...
+profiles {
```
MaxUlysse:
```suggestion
```

PR #146, opened 2020-04-08
Adding config for ICR davros cluster

adrlar:
Tested with sarek run successfully


conf/icr_davros.config
```diff
@@ -0,0 +1,38 @@
...
+ * -------------------------------------------------
+ *  Nextflow nf-core config file for ICR davros HPC
+ * -------------------------------------------------
+ * Defines LSF process executor and singularity
+ * settings.
+ * 
+ */
+params {
+  config_profile_description = "Nextflow nf-core profile for ICR davros HPC"
+  config_profile_contact = "Adrian Larkeryd"
```
apeltzer, line 11: Please add either github handle and/or email here so people can get in touch :-)

conf/icr_davros.config
```diff
@@ -0,0 +1,38 @@
...
+}
+
+singularity {
+  enabled = true
+  runOptions = "--bind /mnt:/mnt --bind /data:/data"
```
apeltzer, line 16: Maybe you want `autoMounts = true ` did that not work for you?

adrlar:
Awesome, thanks for checking - will add docs and look at the suggestions. Reason for naming is that IT is procuring a second, separate cluster at the moment so there will be two versions soon.

apeltzer:
> Awesome, thanks for checking - will add docs and look at the suggestions. Reason for naming is that IT is procuring a second, separate cluster at the moment so there will be two versions soon.

That makes sense then - just wanted to check whether there is a reason for it :-)

adrlar:
Changed up a couple of things according to the suggestions, and did more testing to make sure it works.

adrlar:
I hadn't realised there was a yml file for saying the config was to be tested, have added that now. But I'm afraid of hitting large green buttons with scary text in them ... :)

apeltzer:
Looks good , merged 👍

PR #148, opened 2020-04-09
add viralrecon specific config

MaxUlysse:
cf https://github.com/nf-core/viralrecon/pull/39

PR #149, opened 2020-04-09
Viralreconupdate

MaxUlysse:
I believe you did were right in #148, so I reorganized the files, and did the proper documentation.


drpatelh:
You arrrrreeeee the best @MaxUlysse 😺

PR #150, opened 2020-04-28
update rnafusion munin specific config

MaxUlysse:


MaxUlysse:
making a `genomes.config`, so only need a `params.genomes_base` for the reference, and actually using the `singularity.cacheDir` for the containers, so no need to specify them in a separate config file as they are already specify for docker in the main one

ewels:
Nice!

PR #151, opened 2020-05-04
Add additional custom nf-core/eager params for SHH profile

jfy133:
@TCLamnidis noticed he got very different results with nf-core/eager when mapping, which we identified being stemmed from me setting the default `-n` bwa aln parameter to 0.03 (published in Schubert 2012) for aDNA. However EAGER1 used a `-n` of 0.01, and is likely expected by users in the  department, therefore we will modifiy the default here for SHH.


README.md
```diff
@@ -103,6 +103,7 @@ Currently documentation is available for the following systems:
 * [CRICK](docs/crick.md)
 * [CZBIOHUB_AWS](docs/czbiohub.md)
 * [CZBIOHUB_AWS_HIGHPRIORITY](docs/czbiohub.md)
+* [EBC](docs/ebc.md)
```
apeltzer, line 106: Whats this? ;-)

jfy133:
@apeltzer done!

PR #153, opened 2020-05-22
Add scratch = true to avoid writing of intermediate file to /tmp

FriederikeHanssen:


FriederikeHanssen:
@ewels  we tested it, and it only worked then.
Thanks for the review :)

PR #159, opened 2020-06-11
Add TACC LS5 Config

Emiller88:
----

#


Emiller88:
This should be fine, the thing breaking the CI, is the env variable `$WORK` not being available.

If anyone has a better way to set the `NXF_OPTS` or any other ideas to work around the memory limitations let me know!

.github/workflows/main.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     needs: test_all_profiles
     strategy:
         matrix:
-          profile: ['awsbatch', 'bi','bigpurple', 'binac', 'cbe', 'ccga_dx', 'ccga_med', 'cfc', 'cfc_dev', 'crick', 'denbi_qbic', 'ebc', 'genotoul', 'genouest', 'gis', 'google', 'hebbe', 'icr_davros', 'kraken', 'munin', 'pasteur', 'phoenix', 'prince', 'shh', 'uct_hex', 'uppmax', 'utd_ganymede', 'uzh']
+          profile: ['awsbatch', 'bi','bigpurple', 'binac', 'cbe', 'ccga_dx', 'ccga_med', 'cfc', 'cfc_dev', 'crick', 'denbi_qbic', 'ebc', 'genotoul', 'genouest', 'gis', 'google', 'hebbe', 'icr_davros', 'kraken', 'munin', 'pasteur', 'phoenix', 'prince', 'shh', 'tacc_ls5', 'uct_hex', 'uppmax', 'utd_ganymede', 'uzh']
```
ewels, line 19: This one-line list is getting kind of long...

```suggestion
          profile:
            - awsbatch
            - bi
            - bigpurple
            - binac
            - cbe
            - ccga_dx
            - ccga_med
            - cfc
            - cfc_dev
            - crick
            - denbi_qbic
            - ebc
            - genotoul
            - genouest
            - gis
            - google
            - hebbe
            - icr_davros
            - kraken
            - munin
            - pasteur
            - phoenix
            - prince
            - shh
            - tacc_ls5
            - uct_hex
            - uppmax
            - utd_ganymede
            - uzh
```

conf/tacc_ls5.config
```diff
@@ -0,0 +1,36 @@
+singularityDir = "$WORK/singularity_cache"
+
+params {
+  config_profile_description = 'The Lonestar 5 (LS5) system TACC HPC cluster profile provided by nf-core/configs'
+  config_profile_contact = 'Edmund Miller(@emiller88)'
```
ewels, line 5:
```suggestion
  config_profile_contact = 'Edmund Miller (@emiller88)'
```

conf/tacc_ls5.config
```diff
@@ -0,0 +1,36 @@
...
+  config_profile_url = 'https://portal.tacc.utexas.edu/user-guides/lonestar5'
+}
+
+singularity {
+  enabled = true
+  envWhitelist='SINGULARITY_CACHEPATH'
```
ewels, line 11:
```suggestion
  envWhitelist = 'SINGULARITY_CACHEPATH'
```

docs/tacc_ls5.md
```diff
@@ -0,0 +1,24 @@
...
+```bash
+## Singularity environment modules
+module restore
+module load tacc
+```
+
+:warning: Due to the resource limit of users on the login nodes, and not being able to submit jobs from compute nodes you have to call the `NXF_OPTS` in the same line as the nextflow command. All of this even though the ulimit claims 8GB of virtual memory. Setting it in your `.bashrc` causes it to still fail. It may also run out of memory while building the singularity image and you can download it before running. For how to do that see the  [nf-core offline docs](https://nf-co.re/usage/offline)
+
+```bash
+NXF_OPTS='-Xms2G -Xmx2G -Xss228K' nextflow run nf-core/... -profile tacc_ls5
```
ewels, line 18: Can maybe write this as its own `export NXF_OPTS='-Xms2G -Xmx2G -Xss228K'` line instead? Then recommend that people add this to their `.bashrc` file so that they don't have to remember it for every time that they launch Nextflow...

conf/tacc_ls5.config
```diff
@@ -0,0 +1,36 @@
+singularityDir = "$WORK/singularity_cache"
```
ewels, line 1: I'm sure that this has come up somewhere else and been solved. @nf-core/core can anyone remember this?

apeltzer: CFC e.g. uses a separate singularity cachedir for everyone at QBIC:

https://github.com/nf-core/configs/blob/master/conf/cfc.config

Not sure the `$WORK` is a good idea as users would need to have `$WORK` set accordingly before running the workflow

docs/tacc_ls5.md
Emiller88: I tried that and it still failed for whatever reason, because their login nodes are pretty picky. I'm not even sure if this would work with a real amount of samples this was just the sarek test data. 

The admins have decided they don't like nextflow running on the login nodes at all, so I'm reaching out to the life sciences group to see why they decided they didn't want nextflow used on their clusters, because I think the benefit outweighs the cost of having a footprint on the login node. I'll make a note that nextflow isn't supported in the docs, and leave this for future reference.

conf/tacc_ls5.config
Emiller88: I couldn't get it to pickup the `$SINGULARITY_CACHEPATH` variable, when trying to set it there. 

`$WORK` is also set for all users, they have $HOME $WORK and $SCRATCH on tacc for each user.

.github/workflows/main.yml
Emiller88: I was just thinking the same thing, and it led to some difficult diffs.

docs/tacc_ls5.md
ewels: Yeah, sounds like a bit of a tricky situation. You can always remind the admins that they could choose to allow job submission from compute nodes, then everyone would be happy 😄

conf/tacc_ls5.config
ewels: I wonder if you could use a groovy function with a default to avoid the CI errors?

```groovy
def tacc_work_dir = (System.getenv("$WORK") ?: '') + '/singularity_cache'
```

docs/tacc_ls5.md
Emiller88: > allow job submission from compute nodes
I was told using an entire compute node for job submission was a "waste of resources". 

The life sciences group was helpful and said I might be able to do something like:
```
[compute]$ sbatch my_job.slurm
```
instead do:
```
[compute]$ ssh login1 sbatch my_job.slurm
```
Basically if I can get nextflow to put ssh login1 in front of it's sbatch jobs it should work.

conf/tacc_ls5.config
Emiller88: @ewels I tried this but it appears breaking up the one line broke the CI.

docs/tacc_ls5.md
ewels: Can you not allocate a partial node when you use `sbatch`? Just specify a few gigs of memory / few cpus and then share the node with other jobs?

I don't really understand what you mean with the `ssh` thing - I suspect that you'll have problems even if you managed it though, as Nextflow needs to be able to poll the job scheduler to track job statuses. So if `sbatch` doesn't work then the job tracking probably also won't work.

Emiller88: You could definitely only request a small amount of resources. I'm not a slurm expert, so I don't know if they can let other jobs use the rest of the node(Though that would make sense right?).

So you can run sbatch from the compute node, it just fails because you can't submit from anything but the login node. And I'm going to guess nextflow uses sinfo under the hood or some other slurm helper, which those work from the compute nodes. 

Basically the idea would be every time nextflow goes to submit a job and calls `sbatch work/aa/asdsd/.command.run` it would call `ssh login1 sbatch ...` instead, and ssh into the login node, and call submit the sbatch job there, and do all the "computing" on the computer node, working out channels, preparing the next sbatch job, polling the queue, outputting docs, etc.

ewels: Good luck 😃  👍

conf/tacc_ls5.config
ewels: I don't think that it's this code, I think it's my suggested change to move the long list into a multi-line list. The following code:

https://github.com/nf-core/configs/blob/bb7f67e2b556e1deda34b51799d654603d9c8c9f/bin/cchecker.py#L56-L58

This isn't very nice code, I'll see if I can refactor it in your PR here for you.

ewels: Rewritten in 0eac584 to load the YAML instead of using a regex - hopefully the CI tests should now pass...

ewels:
Ok, we're getting somewhere - now the tests are failing for a more sensible reason 😅 

```
N E X T F L O W  ~  version 20.04.1
Launching `/home/runner/work/configs/configs/configtest.nf` [serene_sammet] - revision: 0a89cddca7
Unknown config attribute `WORK` -- check config file: /home/runner/work/configs/configs/nextflow.config
```

ewels:
Note that you can replicate this test locally with the following command (from your nf-core/configs root directory):

```console
$ nextflow run configtest.nf -profile tacc_ls5 --custom_config_base=./
N E X T F L O W  ~  version 20.05.0-edge
Launching `configtest.nf` [happy_fourier] - revision: 0a89cddca7
Unknown config attribute `WORK` -- check config file: /Users/philewels/GitHub/nf-core/configs/nextflow.config
```

jfy133:
Hi @Emiller88  I'm closing this as stale as it's not been updated in over a year. Feel free to re-open as necessary.

PR #160, opened 2020-06-16
Merge czbiohub_aws_highpriority profile into czbiohub_aws

ewels:
Deleted the `czbiohub_aws_highpriority` profile and moved this into the `czbiohub_aws` profile.

The high priority profile only overwrote part of the existing `czbiohub_aws`, so there was no need to keep these two separate. I changed it to use the child-profile approach that we have setup for UPPMAX `devel`. I also updated the CZI docs to describe this.

Note the slight change in nextflow commands after this change:

```diff
- nextflow run -profile czbiohub_aws_highpriority
+ nextflow run -profile czbiohub_aws,highpriority
```

The `highpriority` profile will only exist _after_ the `czbiohub_aws` profile is loaded. So order of arguments is important. But the flip side is that this rather generic profile name is unlikely to clash with other profiles.

Please don't merge until @olgabot has looked over this, as it is her config 😄 

Phil


MaxUlysse:
Thanks @ewels for doing this, I've been meaning to do this since the last Stockholm Hackathon

PR #163, opened 2020-06-18
resolve  #162 bug

### LINKED ISSUE #162: BI config breaking stuff with `nextflow config`
This code block breaks the functionality of `nextflow config -show-profiles` with all nf-core pipelines:

https://github.com/nf-core/configs/blob/bb7f67e2b556e1deda34b51799d654603d9c8c9f/conf/bi.config#L10-L18

Gives:

```console
$ nextflow config -a nf-core/rnaseq
ERROR: Environment variable NXF_GLOBAL_CONFIG is missing. Set it to point to global.config file.
WARNING: Could not load nf-core/config profiles: https://raw.githubusercontent.com/nf-core/configs/master/nfcore_custom.config
No such file: Config file does not exist: https://raw.githubusercontent.com/nf-core/configs/master/conf/conf/test.config
```

I thought that this was a nextflow bug at first, see https://github.com/nextflow-io/nextflow/issues/1639

@apeltzer - what's going on in this code? Can we refactor it somehow?

piotr-faba-ardigen:
This is work in progress. It resolves the problem found by @ewels , but is not exactly the ideal solution for BI

### DIFF
--- a/conf/bi.config
+++ b/conf/bi.config
@@ -3,16 +3,18 @@ params{
   config_profile_contact = 'Alexander Peltzer (@apeltzer)'
   config_profile_url = 'https://www.boehringer-ingelheim.com/'
 }
-params.globalConfig = determine_global_config()
 
-includeConfig params.globalConfig
+params.globalConfig = set_global_config()
 
-def determine_global_config() {
-    if( System.getenv('NXF_GLOBAL_CONFIG') == null)
-    {
-        def errorMessage = "ERROR: Environment variable NXF_GLOBAL_CONFIG is missing. Set it to point to global.config file."
-        System.err.println(errorMessage)
-        throw new Exception(errorMessage)
-    }
-    return System.getenv('NXF_GLOBAL_CONFIG')   
+
+def set_global_config() {
+  def config = System.getenv('NXF_GLOBAL_CONFIG')
+  if(config  == null)
+  {
+      def errorMessage = "WARNING: For bi.config requires NXF_GLOBAL_CONFIG env var to be set. Point it to global.config file if you want to use this profile."
+      System.err.println(errorMessage)
+  }else{
+    includeConfig config
+  }
+  return config
 }


ewels:
Yes looks great! Seems to me that the functionality would be identical? Or what am I missing?

apeltzer:
> Yes looks great! Seems to me that the functionality would be identical? Or what am I missing?

Seems to me the same :-)

piotr-faba-ardigen:
@apeltzer I got distracted yesterday and didn't submit the comment. This resolves the `-show-profiles` problem. But, since there is no exception raised, user will not see exactly why his pipeline fails. If someone will not have have NXF_GLOBAL_CONFIG set, the pipeline will start and fail when it will try to run docker instead of singularity containers. The issue is that the warning is at the top of the pipeline, so it's not directly visible to the user as the root cause for pipeline failure.

I will submit something else if I get any ideas. This solution is a compromise for both sides, so it can stay for now.

ewels:
Yes, this desire to print warnings at the end of execution is fairly common. I’ve double-printed a few things like this in the past for the same reason (start and end of log).

How about if we have a `params` variable common to all of nf-core that can have a list of warning messages to print at the end of every pipeline run? Then you could just append to that list here if there’s a problem..

piotr-faba-ardigen:
I think params variable for appending warnings would be great. I use that in our pipelins for communicating warnings regarding not using particular parameter that despite being set by the user is not used in particular context. 

The alternative with making the Exception hostname dependent is probably not gonna fly on our end - too many hosts that are in use.

piotr-faba-ardigen:
I think params variable for appending warnings would be great. I use that in our pipelines for communicating warnings regarding not using particular parameter that despite being set by the user is not used in particular context. 

The alternative with making the Exception hostname dependent is probably not gonna fly on our end - too many hosts that are in use.

PR #170, opened 2020-07-07
Make walltimes more sophisticated and add a HOPS profile for nf-core/eager @ SHH

jfy133:
This update has two different additions:

* Adds a `hops` profile ofr the Pathogen group(s) for their default initial screening of everything coming off their sequencers
* It makes more a sophisticated trey system for all EAGER processes, to account for large sequencing data (whole HiSeq lane-scale). Rather than just adding an extra 2 hours, it will now switch to each next larger queue, on each subsequent retry.


conf/pipeline/eager/shh.config
```diff
@@ -22,6 +22,44 @@ process {
 }
 
 profiles {
+  hops {
+    pathogen_screening {
+      fasta = '/projects1/Reference_Genomes/Human/HG19/hg19_complete.fasta'
```
ewels, line 27: I think you're missing the `params` scope here..

PR #178, opened 2020-09-03
Add Max Planck Computing and Data Facility Cobra/Raven Clusters

jfy133:


docs/mpcdf.md
```diff
@@ -0,0 +1,45 @@
...
+
+To run Nextflow, the `jdk` module must be loaded. To use the nf-core profile(s), run the pipeline with `-profile <cluster>,mpcdf`.
+
+Currently the following clusters are supported: cobra, raven
+
+>NB: Nextflow will need to submit the jobs via SLURM to the clusters and as such the commands above will have to be executed on one of the head nodes. Check the [MPCDF documentation](https://www.mpcdf.mpg.de/services/computing).
+
+## cobra
+
+Cobra does not currently support singularity, therefore the anaconda/module is loaded for each process.
```
apeltzer, line 15: 🤦‍♂️ Might be problematic with e.g. DSLv2 pipelines that heavily rely on biocontainer(s).

apeltzer (approved):
Good - might just be an issue with conda and DSLv2 pipelines soon as they rely on biocontainers heavily :-(

docs/mpcdf.md
jfy133: Yup, don't worry. They aren't against installing it later, (also I've not added the older draco cluster which does support singularity), just want more evidnce of usage first.

PR #179, opened 2020-09-14
feat: add Kim queue

Emiller88:
----

Added a new node that's big enough to run STAR with a human genome.

This also bumps the singularity version to 3.2.1 because it appears to be working for me so far.

Double check my queue logic please though. Going for

Greater than 32GB memory and CPU are less than or equal to 12: Kim

Less than 24GB memory and CPU are less than or equal to 8: smallmem

else genomics

### DIFF
--- a/conf/utd_ganymede.config
+++ b/conf/utd_ganymede.config
@@ -14,7 +14,7 @@ singularity {
 process {
   beforeScript = 'module load singularity/3.2.1'
   executor = 'slurm'
-  queue = { task.memory >= 32.GB ? 'Kim': task.memory <= 24.GB && task.cpu <= 8 ? 'smallmem' : 'genomics' }
+  queue = { task.memory >= 32.GB && task.cpu <= 12 ? 'Kim': task.memory <= 24.GB && task.cpu <= 8 ? 'smallmem' : 'genomics' }
 }
 
 params {
--- a/conf/utd_ganymede.config
+++ b/conf/utd_ganymede.config
@@ -12,13 +12,13 @@ singularity {
 }
 
 process {
-  beforeScript = 'module load singularity/2.4.5'
+  beforeScript = 'module load singularity/3.2.1'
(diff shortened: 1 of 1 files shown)


jfy133 (approved):

drpatelh:
Just a minor point. You don't strictly speaking need to load Singularity using `beforeScript` because it has to be loaded before you run the pipeline, and  hence should be inherited in the job submission environment. It doesn't do any harm but saves you from having to update the version periodically.

jfy133:
> Just a minor point. You don't strictly speaking need to load Singularity using `beforeScript` because it has to be loaded before you run the pipeline, and hence should be inherited in the job submission environment. It doesn't do any harm but saves you from having to update the version periodically.

Good to know - although last week I came across a context where the singularity module couldn't be loaded on the headnode (banned by sysadmins for some reason), so `beforeScript` is useful in those cases. Just posting here (my personal) future reference ;).

drpatelh:
Yep, you are absolutely right 👍 Does depend on how the cluster is set up so can be tested beforehand to see if everything works without it.

Emiller88:
Thanks for the feedback!

The version bump was due to https://github.com/nf-core/configs/pull/139. I expect them to keep this module so hopefully I won't have to change the version in the future.

I did some testing on the `beforeScript` because I had noticed some different behaviors over time.

So what I found. On the first run of a new pipeline, you have to `module load singularity` on the head node, because the way nf-core works currently(or so I understand) it uses one big singularity image, and the head node has to download that(https://github.com/nf-core/configs/pull/159 had issues with it here too). Once that image is cached though, you don't have to load singularity on the head node, and the `beforeScript` will take care of loading them on the compute nodes. So I'll leave it just in case users forget to load the singularity module. As @drpatelh mentioned though loading the singularity module on the head node works and the compute nodes inherit the environment.

jfy133:
That matches my own experience!

PR #185, opened 2020-11-09
awsbatch config update

ggabernet:
- Updates to allow pipelines resume when running on AWS batch.


ewels (approved):
LGTM 🚀

ewels:
x-ref https://github.com/nextflow-io/nextflow/issues/813#issuecomment-724124512

Note that this new feature presumably isn't released yet. I'm not sure if adding this config before it's generally available will break stuff?

ggabernet:
yes, that's right. We wait for the next Nextflow release before merging 👍

ewels:
Maybe worth also adding the minimum nextflow version to this config? Or perhaps that can only be put in the main pipeline? Not sure..

apeltzer (approved):
I wondered why there is no issue open about this, but I've only opened one internally here to fix this 😆 Good catch!

ewels:
Happy to merge this now that the edge release is out (and we are requiring that for many pipelines / Tower always runs it etc).

PR #198, opened 2020-12-18
[WIP] Add Institute-specific genomes!!

drpatelh:
I have now tested this with the rnaseq pipeline (v3.0) on my own fork. This means we can have remote genome configs for each Institution so all users can just provide a `--genome` key from their own `genomes.config` and the files can be shared across all users saving the need to duplicate the data across your storage!!!

Since this file is on nf-core/configs if you have/add your Institutional profile here then these configs will be pulled dynamically when you run any nf-core pipeline meaning that if you update the genomes file here then that will be observed immediately for all users too i.e. saving the need to change this manually across multiple users. 

Still fascinated that this repo is still working since we first put this together a couple of years ago. Being able to use remote genomes is now a massive bonus too!


conf/crick.config
```diff
@@ -1,23 +1,32 @@
...
 }
 
 process {
   executor = 'slurm'
+
+  withName:RSEM_CALCULATEEXPRESSION  {
+    cpus   = 12
+    memory = 72.GB
+    time   = 16.h
+  }
```
MaxUlysse, line 23: This part seems to be pipeline specific.
Should be in a special `rnaseq/crick.config` file, no?

drpatelh: Genius @MaxUlysse !

conf/genomes/crick/genomes.config
```diff
@@ -0,0 +1,51 @@
...
+ * nextflow run nf-core/rnaseq \
+ *     --input samplesheet.csv \
+ *     --genome GRCh38.95.CAMP \
+ *     -profile crick \
+ */
+
+params {
+  genomes {
+    'GRCh38.95.CAMP' {
+      fasta       = "/camp/svc/reference/Genomics/babs/homo_sapiens/ensembl/GRCh38/release-95/genome/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa"
```
MaxUlysse, line 15: You could add a `params.genomes_base` for `/camp/svc/reference/Genomics/babs` to simplify the setup and make it easier to steal for other people ;-)

MaxUlysse:
I do like this idea.

I think you're only missing a bit more information in the `crick.md` file, like on the CAMP specific genomes ;-)

conf/crick.config
drpatelh: Done in https://github.com/nf-core/configs/pull/198/commits/49d18f4fcccd50b2d27e38830afafd41834ac021 and  https://github.com/nf-core/configs/pull/198/commits/1ae378b65cdb787877d6351906b78ac0b6c5a429

conf/genomes/crick/genomes.config
drpatelh: I did think about doing that but I figured that I would try and leave it as vanilla as possible for now. Also, having explicit paths is much easier to read and update. Would still need to create your own custom configs with these paths if you wanted to steal anyway! :)

drpatelh:
We will be adding more genomes in the future @MaxUlysse . I mainly want to get this in so users can test it properly. Will definitely add docs then 👍🏽 Also move the pipeline specific params as you suggested.

MaxUlysse:
> We will be adding more genomes in the future @MaxUlysse . I mainly want to get this in so users can test it properly. Will definitely add docs then 👍🏽 Also move the pipeline specific params as you suggested.

I'm afraid you also need to add the specific part in the rnaseq pipeline
cf https://github.com/nf-core/configs/#adding-a-new-pipeline-specific-config.

So you can actually keep it the way it was until you make a new release ;-)

drpatelh:
Ah dammit! Good spot. Ok, will revert it back.

drpatelh:
Done @MaxUlysse ! This custom rnaseq stuff has already been fixed for the next release anyway :)

ewels:
@drpatelh / @chris-cheshire - is this PR still alive?

drpatelh:
POC so will convert to draft.

jfy133:
@drpatelh Given this is 4 years old, I'm leaning towards closing it and asking oyu to start from scrach.

Will give you a week to think about it!

maxulysse:
> @drpatelh Given this is 4 years old, I'm leaning towards closing it and asking oyu to start from scrach.
> 
> Will give you a week to think about it!

I'd say we kill it

jfy133:
Not heard back otherwise, closing!

PR #199, opened 2021-01-15
added anaconda module for cbe

rbpisupati:
Module for anaconda was added to cbe


phue (approved):
All good, nice to have this as a fallback

Emiller88:
@rbpisupati Does the `module =` load a module on the cluster like `module load anaconda3`?

rbpisupati:
yes right! it would be doing that,.

PR #202, opened 2021-01-26
Add debug profile

TCLamnidis:
which deactivates cleanup of workdir after successful run. Useful for debugging and testing of code changes, as it is possible to see exactly what the process did on successful runs.


jfy133:
Lgtm!

jfy133 (approved):

PR #210, opened 2021-03-11
Cfc sarek

FriederikeHanssen:
I now added the max values used for the biggest input data I had. They are probably still a bit over the top for some processes, but much less then the values used before.

Tested this by supplying the config manually.

### DIFF
--- a/conf/pipeline/sarek/cfc.config
+++ b/conf/pipeline/sarek/cfc.config
@@ -19,7 +19,7 @@ process {
     memory = 372.GB
   }
   withName:MapReads{
-    cpus = 48
+    cpus = 20
     memory = 128.GB
   }
 }
--- a/conf/pipeline/sarek/cfc.config
+++ b/conf/pipeline/sarek/cfc.config
@@ -1,0 +1,24 @@
+// Profile config names for nf-core/configs
+
+params {
+  // Specific nf-core/configs params
+  config_profile_contact = 'Friederike Hanssen (@FriederikeHanssen'
+  config_profile_description = 'nf-core/sarek CFC profile provided by nf-core/configs'
+}
+
+// Specific nf-core/sarek process configuration
+process {
+  withName:'StrelkaSingle|Strelka|StrelkaBP|MantaSingle|Manta' {
+    memory = {check_resource(128.GB)}
+    cpus = 16
+  }
+  withName:'MSIsensor_scan|MSIsensor_msi' {
+    memory = {check_resource(128.GB)}
+  }
+  withName:BamQC {
+    memory = {check_resource(372.GB)}
+  }
+  withName:MapReads{
+    cpus = 48
+  }
+}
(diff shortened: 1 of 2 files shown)


conf/pipeline/sarek/cfc.config
```diff
@@ -0,0 +1,24 @@
...
+params {
+  // Specific nf-core/configs params
+  config_profile_contact = 'Friederike Hanssen (@FriederikeHanssen)'
+  config_profile_description = 'nf-core/sarek CFC profile provided by nf-core/configs'
+}
+
+// Specific nf-core/sarek process configuration
+process {
+  withName:'StrelkaSingle|Strelka|StrelkaBP|MantaSingle|Manta' {
+    memory = 128.GB
```
ggabernet, line 12: I so much memory necessary? Basically anything over 60.GB will go to `qbic` queue and occupy the high-mem nodes

conf/pipeline/sarek/cfc.config
```diff
@@ -0,0 +1,24 @@
...
+    cpus = 16
+  }
+  withName:'MSIsensor_scan|MSIsensor_msi' {
+    memory = 128.GB
+  }
+  withName:BamQC {
+    memory = 372.GB
+  }
+  withName:MapReads{
+    cpus = 48
```
ggabernet, line 22: Also here this will go to qbic queue. Which is fine for the Mapping step I guess. But then also more memory could be used to just 60.GB?

conf/pipeline/sarek/cfc.config
FriederikeHanssen: Fair enough. For an initial attempt I reduced anything that was labeled with `max_cpus/memory` to whatever max setting I had. I think, we can risk going much lower and just increase, if too many people report failures. I'd be happy to keep taps on that and increase if needed.

conf/pipeline/sarek/cfc.config
FriederikeHanssen: so far we haven't had issues with 'just' using 60GB memory. Looking at the execution_reports for my large data sets, I supplied 372GB but then only 120GB or so were used. I don't think it makes sense with those numbers to go over 128Gb or so. (Could change for BWAMem2, that seems to scale better with increased resource, but that is a thing for the next release).

conf/pipeline/sarek/cfc.config
FriederikeHanssen: So for now maybe everything to CFC queue with 59Gb of memory and 19 cpus and then evaluate over a few weeks whether any significant problems or delays occur?

ggabernet: that sounds good, thanks!

ggabernet:
```suggestion
    memory = 59.GB
```

conf/pipeline/sarek/cfc.config
ggabernet: ok if "just" 60GB are needed then we can use again the `compute` queue, where nodes have 20 cpus. Would that be ok?

ggabernet, line 22:
```suggestion
    cpus = 20
```

FriederikeHanssen: Sure, that way we keep it really more on the lower end. I will then also reduce the memory to 60GB. Since `retry` now works, for people with bigger datasets it should automatically resubmit with double the amount then.

ggabernet (approved):
LGTM

PR #215, opened 2021-03-22
Update Ganymede Config

Emiller88:
Been meaning to do this for awhile. From looking the some of the eager configs it looks like check_max works out of the box now.

### DIFF
--- a/conf/utd_ganymede.config
+++ b/conf/utd_ganymede.config
@@ -5,6 +5,11 @@ params {
   config_profile_url = 'http://docs.oithpc.utdallas.edu/'
 }
 
+env {
+    TMPDIR = '/home/$USER/scratch/tmp'
+    SINGULARITY_CACHEDIR = '/home/$USER/scratch/tmp'
+}
+
 singularity {
   enabled = true
   envWhitelist='SINGULARITY_BINDPATH'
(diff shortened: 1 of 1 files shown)


conf/utd_ganymede.config
```diff
@@ -15,6 +20,28 @@ process {
...
+    time = { check_max( 6.h * task.attempt, 'time' ) }
+  }
+
+  withLabel:process_medium {
+    cpus = { check_max( 16 * task.attempt, 'cpus' ) }
+    memory = { check_max( 31.GB * task.attempt, 'memory' ) }
+    time = { check_max( 8.h * task.attempt, 'time' ) }
+  }
+
+  withLabel:process_large {
```
drpatelh, line 40: This label doesn't exist in the nf-core pipeline template? Is it a custom one? Are the other default `label` breaking for you?

Emiller88: Ah you're right it should be `high`. Those we're getting the right queue anyways, the main concern was bumping down the medium process to the specs that we have more nodes on.

ewels (approved):
Bear in mind that using a tmpdir for singularity means re-downloading the images instead of reusing. But I guess that you realise this already..

PR #220, opened 2021-03-26
UPPMAX: Avoid undefined parameter warnings

ewels:
Fixes error without needing the parameter to be added to the pipeline config:

```
WARN: Access to undefined parameter `project` -- Initialise it to a default value eg. `params.project = some_value
```

Adds param names to `schema_ignore_params` so that theoretically they do not trigger a linting warning:

```
WARN: Found unexpected parameters:
* --cluster-options: null
* --clusterOptions: null
* --project: snic2017-7-196
- Ignore this warning: params.schema_ignore_params = "cluster-options,clusterOptions,project"
```

In my testing everything still works as expected, but the lint warnings are still triggered 🤔 Not sure if that's just a cache problem..?

### DIFF
--- a/conf/uppmax.config
+++ b/conf/uppmax.config
@@ -1,8 +1,10 @@
-//Profile config names for nf-core/configs
+// Profile config names for nf-core/configs
 params {
   config_profile_description = 'Swedish UPPMAX cluster profile provided by nf-core/configs.'
   config_profile_contact = 'Phil Ewels (@ewels)'
   config_profile_url = 'https://www.uppmax.uu.se/'
+  project = null
+  clusterOptions = null
 }
 
 singularity {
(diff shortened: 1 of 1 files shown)


MaxUlysse (approved):

conf/uppmax.config
```diff
@@ -1,8 +1,11 @@
-//Profile config names for nf-core/configs
+// Profile config names for nf-core/configs
 params {
   config_profile_description = 'Swedish UPPMAX cluster profile provided by nf-core/configs.'
   config_profile_contact = 'Phil Ewels (@ewels)'
   config_profile_url = 'https://www.uppmax.uu.se/'
+  project = null
+  clusterOptions = null
+  schema_ignore_params = "genomes,input_paths,cluster-options,clusterOptions,project"
```
MaxUlysse, line 8: I don't think `genomes` and `input_paths` should be included here, but I like that

ewels: The problem with `schema_ignore_params` is that it's just a string, so we have to overwrite it and can't just append to it as it stands. So these are here because they're there in the template and I didn't want to overwrite them.

PR #226, opened 2021-04-08
UPPMAX: Allow jobs needing more memory than thin nodes, pick up cluster info through slurm

pontus:
Use slurm to find a node from the correct cluster rather than relying on hostname (support case of login nodes for rackham being used for snowy and snowy being accessed with `SLURM_CLUSTERS=snowy`).

Raise `max_memory` to the largest node on the specific cluster.

Add handling to modify the sbatch directives as needed if more memory is required than is supported on a single node.

See also #220.

I've tried verifying this as far as I can, but that is a fairly limited degree, so both scrutiny and practical testing would be very welcome.

### DIFF
--- a/conf/uppmax.config
+++ b/conf/uppmax.config
@@ -9,7 +9,7 @@ singularity {
   enabled = true
 }
 
-def hostname = "hostname".execute().text.trim()
+def hostname = "sinfo --local -N -h | head -1 | cut -f1 -d' ' ".execute().text.trim()
 
 process {
   // closure to create a suitable clusterOptions
--- a/conf/uppmax.config
+++ b/conf/uppmax.config
@@ -9,25 +9,48 @@ singularity {
   enabled = true
 }
 
+def hostname = "hostname".execute().text.trim()
+
 process {
+  // closure to create a suitable clusterOptions
+  def clusterOptionsCreator = { m ->
+                            String base = "-A $params.project ${params.clusterOptions ?: ''}"
+			    // Do not use -p node on irma or if a thin node/core is enough
+			    if (m < 125.GB || hostname ==~ "i.*") {
+			      return base
+			    }
+			    if (m < 250.GB) {
+			      return base + " -p node -C mem256GB "
+			    }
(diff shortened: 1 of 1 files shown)


conf/uppmax.config
```diff
@@ -9,25 +9,48 @@ singularity {
   enabled = true
 }
 
+def hostname = "sinfo --local -N -h | head -1 | cut -f1 -d' ' ".execute().text.trim()
+
 process {
+  // closure to create a suitable clusterOptions
+  def clusterOptionsCreator = { m ->
```
ewels, line 16: I'd be inclined to keep this out of the `process` scope, alongside `hostname` instead

conf/uppmax.config
```diff
@@ -9,25 +9,48 @@ singularity {
...
+if (hostname ==~ "b.*") {
+  params.max_memory = 500.GB
+}
 
-if (hostname ==~ "b.*" || hostname ==~ "s.*") {
-  params.max_memory = 109.GB
+if (hostname ==~ "s.*") {
+  params.max_memory = 500.GB
+  params.max_time = 700.h
 }
```
ewels, line 54: The `sinfo` thing is nice, but do we actually need to differentiate between snowy and rackham? 😅 They seem to be using basically the same config, only the `max_time` that's different (not sure what the history of that is)

ewels (commented):
Super nice, it's been in the back of my mind to attempt something like this for ages - thank you!

Couple of comments, might try to whip together a minimal pipeline example to do some testing with later.

conf/uppmax.config
pontus: I think differentation is useful because of memory differences; memory per core could probably be calculated (?) but the maximum node size and corresponding constraints will differ.

(Node jobs can have runtimes up to 30 days on snowy.)

conf/uppmax.config
pontus: Thanks, fixed.

conf/uppmax.config
pontus: (Wrapping in a try/catch seems to handle the `sinfo` in test issue nicely.)

PR #227, opened 2021-04-08
Added microscope profile

TCLamnidis:
Added profile definition for MICROSCOPE project. will update profile as needed for the project.


jfy133 (changes requested):
Please add a note to the pipeline specific documentation and the GTG

TCLamnidis:
After discussions with James, we agreed that adding a  project specific profile here was not the right approach.

PR #232, opened 2021-04-15
Update mpcdf to use singularity

jfy133:
`raven` and `cobra` now support singularity. This switches the default containers for both accordingly.


apeltzer (approved):
Much better and reliable ;-)

PR #238, opened 2021-04-29
Fix hebbe withName regex

ewels:
Turns out that this syntax doesn't work (thanks for testing @BarryDigby!). I don't think anyone uses this config profile as no-one has told me this in the 3+ years since I wrote this code, but still it would be nice if the code is valid for future reference..


pontus (approved):
Looks legit, untested.

PR #244, opened 2021-05-25
Add TMPDIR and mount for it to singularity configuration + Migrate to BioinformaticsResources for genomes.

oalmelid:
This config update is intended to affect only the University of Edinburgh eddie cluster configuration.
It fixes an issue with /tmp filling up when using singularity because:

- TMPDIR isn't set inside the container environment
- mktemp therefore defaults to /tmp inside the container
- Singularity on eddie has been configured to mount /tmp inside the container

/tmp is tiny on eddie nodes, so this causes issues.

It also migrates ot using BioinformaticsResources rather than the deprecated NextGenResources.

### DIFF
--- a/conf/eddie.config
+++ b/conf/eddie.config
@@ -31,7 +31,7 @@ process {
 params {
   saveReference = true
   // iGenomes reference base
-  igenomes_base = '/exports/igmm/eddie/NextGenResources/igenomes'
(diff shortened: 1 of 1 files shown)


oalmelid:
This has been tested with the nf-core/chipseq pipeline on eddie, though no other pipelines have been tested.

PR #246, opened 2021-06-04
Make eager@SHH big_data go by default to medium via run time

jfy133:
As discussed on mattermost


TCLamnidis (approved):
Why run twice with 48h though?

jfy133:
you would still get increase CPUs/mem?

PR #249, opened 2021-06-10
Update cfc.config

mseybold:
singularity is now installed natively on the nodes, so no more need for a module here


ggabernet (approved):
Thanks, Matze, looks good to me!

PR #252, opened 2021-06-17
Add UTD sysbio

Emiller88:


conf/utd_sysbio.config
```diff
@@ -0,0 +1,43 @@
...
+  cacheDir = params.singularity_cache_dir
+}
+
+process {
+  beforeScript = 'module load singularity/3.4.1'
+  executor = 'slurm'
+  queue = { task.memory >= 30.GB && task.cpu <= 16 ? 'normal': 'smallmem' }
+
+
+  withName:STAR_ALIGN {
```
jfy133, line 26: This should go into a pipeline specific config file: https://github.com/nf-core/configs/tree/master/conf/pipeline/rnaseq

conf/utd_sysbio.config
```diff
@@ -0,0 +1,43 @@
...
+
+env {
+    TMPDIR = '/home/$USER/scratch/tmp'
+}
+
+singularity {
+  enabled = true
+  envWhitelist='SINGULARITY_BINDPATH'
+  autoMounts = true
+  cacheDir = params.singularity_cache_dir
```
jfy133, line 17: This should maybe go in a pipeline specific config? Unless it's a new DSL2 template feature (I've not dived into that yet 😅)

conf/utd_sysbio.config
```diff
@@ -0,0 +1,43 @@
...
+
+  withName:STAR_ALIGN {
+    memory = 36.GB
+  }
+
+  withLabel:process_high {
+    cpus = { check_max( 16 * task.attempt, 'cpus' ) }
+    memory = { check_max( 60.GB * task.attempt, 'memory' ) }
+    time = { check_max( 16.h * task.attempt, 'time' ) }
+  }
```
jfy133, line 34: This should go into a pipeline specific config file - the labels may not be consistent across all pipelines in regards to what 'e.g.' may refer to.

docs/utd_sysbio.config
```diff
@@ -0,0 +1,18 @@
...
+
+Before running the pipeline you will need to load Singularity using the environment module system on Sysbio. You can do this by issuing the commands below:
+
+```bash
+## Singularity environment modules
+module purge
+module load singularity
+```
+
+All of the intermediate files required to run the pipeline will be stored in the `work/` directory. It is recommended to delete this directory after the pipeline has finished successfully because it can get quite large, and all of the main output files will be saved in the `results/` directory anyway.
```
jfy133, line 15: You could also consider adding `cleanup = true` to empty `work/` of files by default at the end of a successful run, if this could be an issue.

jfy133 (changes requested):
LGTM overall, just need to push some things to a pipeline specific config.

conf/utd_sysbio.config
Emiller88: The singularity cache? Nah, that should work with dsl1 as well. Ideally it should allow everyone to use the same images, so we don't each have our own copies of them. In theory ;)

conf/pipeline/rnaseq/utd_sysbio.config
```diff
@@ -0,0 +1,13 @@
+process {
```
jfy133, line 1: Ah also:

I would suggest also adding the whole `config_description` parameter stuff as well to the pipeline specific config, to make sure users known that there is an underlying change from the `RNASEQ` default _specifically_ for that pipeline a la

https://github.com/nf-core/configs/blob/9cca0eaf85e0ba12835c2cdd62fba12c643fec4a/conf/pipeline/eager/shh.config#L3-L8

conf/utd_sysbio.config
jfy133: Are you sure? I don't see it in the template [`nextflow.config`](https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/nextflow.config) on neither `master` or `dev`, so I would assume it's something specific to a given pipeline? I've never seen it before anyway :grimacing:

jfy133 (changes requested):
Almost there:

You'll need to add the corresponding `docs/pipeline/` file (e.g. https://github.com/nf-core/configs/tree/master/docs/pipeline/rnafusion)

Basically everything described here: https://github.com/nf-core/configs#adding-a-new-pipeline-specific-config (you can ignore 'enabling in the pipeline' section, I guess, but everything else I think is relevant)

maxulysse (approved):
LGTM

conf/pipeline/rnaseq/utd_sysbio.config
```diff
@@ -0,0 +1,19 @@
+params {
+  config_profile_description = 'University of Texas at Dallas HPC cluster profile provided by nf-core/configs'
```
jfy133, line 2: Would suggest specifying is rna-seq specific one but otherwise gtg

jfy133 (approved):
Yay for github mobile!

conf/pipeline/rnaseq/utd_sysbio.config
Emiller88: I'll address this in another PR

maxulysse (approved):

PR #259, opened 2021-07-23
Immcantation profile

ggabernet:
Hi @ssnn-airr ,

this would be the config file. Do you think it makes sense to also add your Slurm cluster settings into it already for other members of your lab?


jfy133:
@ggabernet @ssnn-airr  is this still needed? Or can we close it?

PR #271, opened 2021-09-08
Adds Rosalind config for KCL!

theoportlock:
Followed the steps outlined in the README.md
Hopefully KCL will add a Nextflow module in the near future but this works for now

### DIFF
--- a/conf/rosalind.config
+++ b/conf/rosalind.config
@@ -1,0 +1,28 @@
+params {
+	config_profile_description = 'Kings College London Rosalind HPC'
+	config_profile_contact = 'Theo Portlock'
+	config_profile_url = 'https://www.rosalind.kcl.ac.uk/'
+}
+
+singularity {
+	enabled = true
+	autoMounts = true
+	docker.enabled = false
+}
+
+params {
+	max_memory = 64.GB
+	max_cpus = 16
+	max_time = 24.h
+	partition = 'shared'
+}
+
+process {
+	executor = 'slurm'
+	maxRetries = 3
+	clusterOptions = { "--partition=$params.partition ${params.clusterOptions ?: ''}" }
+}
+
+executor {
+	submitRateLimit = '1 sec'
+}
--- a/conf/rosalind.config
+++ b/conf/rosalind.config
@@ -1,28 +1,29 @@
 params {
-	config_profile_description = 'Kings College London Rosalind HPC'
-	config_profile_contact = 'Theo Portlock'
-	config_profile_url = 'https://rosalind.kcl.ac.uk/'
+  config_profile_description = 'Kings College London Rosalind HPC'
+  config_profile_contact = 'Theo Portlock'
+  config_profile_url = 'https://www.rosalind.kcl.ac.uk/'
 }
 
 singularity {
-	enabled = true
-	autoMounts = true
-	docker.enabled = false
+  enabled = true
+  autoMounts = true
+  docker.enabled = false
 }
 
 params {
-	max_memory = 64.GB
-	max_cpus = 16
-	max_time = 24.h
-	partition = 'shared'
+  max_memory = 64.GB
+  max_cpus = 16
+  max_time = 24.h
+  partition = 'shared'
+  schema_ignore_params = 'partition'
 }
 
 process {
-	executor = 'slurm'
-	maxRetries = 3
-	clusterOptions = { "--partition=$params.partition" }
+  executor = 'slurm'
+  maxRetries = 3
+  clusterOptions = { "--partition=$params.partition" }
 }
 
 executor {
-	submitRateLimit = '1 sec'
+  submitRateLimit = '1 sec'
 }
--- a/conf/rosalind.config
+++ b/conf/rosalind.config
@@ -15,7 +15,7 @@ params {
   max_cpus = 16
   max_time = 24.h
   partition = 'shared'
-  schema_ignore_params = 'partition'
+  schema_ignore_params = 'partition,genomes,modules'
 }
 
 process {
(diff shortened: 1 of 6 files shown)


conf/rosalind.config
```diff
@@ -0,0 +1,28 @@
+params {
+	config_profile_description = 'Kings College London Rosalind HPC'
+	config_profile_contact = 'Theo Portlock'
+	config_profile_url = 'https://www.rosalind.kcl.ac.uk/'
```
jfy133, line 4:
```suggestion
	config_profile_url = 'https://rosalind.kcl.ac.uk/'
```

conf/rosalind.config
```diff
@@ -0,0 +1,28 @@
...
+}
+
+singularity {
+	enabled = true
+	autoMounts = true
+	docker.enabled = false
+}
+
+params {
+	max_memory = 64.GB
```
jfy133, line 14: According to the docs, max memory would be `384.GB`, no?

https://rosalind.kcl.ac.uk/hpc/compute_nodes/

conf/rosalind.config
```diff
@@ -0,0 +1,28 @@
...
+	max_cpus = 16
```
jfy133, line 15: Same again, 32?

https://rosalind.kcl.ac.uk/hpc/compute_nodes/

conf/rosalind.config
```diff
@@ -0,0 +1,28 @@
...
+	max_time = 24.h
+	partition = 'shared'
+}
+
+process {
+	executor = 'slurm'
+	maxRetries = 3
+	clusterOptions = { "--partition=$params.partition ${params.clusterOptions ?: ''}" }
```
jfy133, line 23: The condition there looks a bit broken, is  the `?` in the right place?

jfy133 (commented):
A few comments

conf/rosalind.config
theoportlock: Good suggestion, change made

conf/rosalind.config
theoportlock: I was under the impression that the 10 percent of total resource usage was their policy (https://rosalind.kcl.ac.uk/hpc/scheduler_policy/). I've just tested a job requiring 384.GB and am recieving the "Memory specification can not be satisfied" sbatch error. I have contacted a member of the Rosalind support team for further clarification and will update you when they reply. Thanks for the suggestion!

jfy133: Ah sorry, I thought that was across the entire cluster - I must have misunderstood. What you put makes sense then :+1:

conf/rosalind.config
theoportlock: Good spot, I have removed the broken clusterOptions and replaced with just the partition specification. I have tested this solution also and it works as expected

jfy133: The only issue  I see here is that it would maybe require all nf-core pipelines to a `$params.partition` defined in their JSON, but I'm not 100% sure (i gues sif you tested it, it should work - have you tried different nf-core pipelines). It would be very important to document this, if you can't define it dynamically based on resource requirements.

jfy133 (approved):

jfy133:
@theoportlock I've approved, but it might be wort hasking on #configs on slack whether the `params.partition` approach is OK or not

conf/rosalind.config
```diff
@@ -0,0 +1,28 @@
+params {
+	config_profile_description = 'Kings College London Rosalind HPC'
```
drpatelh, line 2: I think we have used 2 space tabs everywhere else. Be good to keep it consistent :)

conf/rosalind.config
```diff
@@ -0,0 +1,28 @@
...
+	autoMounts = true
+	docker.enabled = false
+}
+
+params {
+	max_memory = 64.GB
+	max_cpus = 16
+	max_time = 24.h
+	partition = 'shared'
+}
```
drpatelh, line 18:
```suggestion
	partition = 'shared'
	schema_ignore_params = 'partition'
}
```

drpatelh: I think this may deal with a custom `partition` parameter being provided here @jfy133 . Be great if you could maybe test the final config with the latest version of nf-core/rnaseq to make sure it works @theoportlock :)

drpatelh (changes requested):
Couple of tiny tweaks otherwise looks awesome @theoportlock 🚀

conf/rosalind.config
theoportlock: Sure thing, I will make those changes now :)

conf/rosalind.config
jfy133: You are totally right and I went through the same question/finding the answer the first time and  I totally forgot 🤦‍♀

conf/rosalind.config
theoportlock: Exactly, I have now ran multiple partitions with multiple nf-core pipelines including rnaseq without error. I also updated the README about the default partition and how to specify other partitions.

conf/rosalind.config
theoportlock: I have tested the final config with rnaseq and other nf-core pipelines on mutliple partitions with your suggested change and they complete without error. Thanks again!

conf/rosalind.config
```diff
@@ -0,0 +1,29 @@
...
+  autoMounts = true
+  docker.enabled = false
+}
+
+params {
+  max_memory = 64.GB
+  max_cpus = 16
+  max_time = 24.h
+  partition = 'shared'
+  schema_ignore_params = 'partition'
```
drpatelh, line 18:
```suggestion
  schema_ignore_params = 'partition,genomes,modules'
```

drpatelh (approved):

drpatelh:
Great! Pushed a commit to update in-line with this PR https://github.com/nf-core/configs/pull/278

PR #277, opened 2021-09-15
Computerome

marcmtk:
Adding an institutional profile for [Computerome](https://www.computerome.dk/) a Danish supercomputer for life sciences. 

I've followed the guide from [Bytesize 10](https://nf-co.re/events/2021/bytesize-10-institutional-profiles) and have tested using nf-core/viralrecon with either of `-c` or `-profile computerome --custom_config_base` syntax.

As computerome requires PBS -A and -W tags to be set i've followed the example of the uppmax configuration file and introduced a params.project. Jobs will fail if it is not set, but they only fail upon first submitting a job to the scheduler. It would be nice if it could fail faster, but i don't know how that would be done.

### DIFF
--- a/conf/computerome.config
+++ b/conf/computerome.config
@@ -1,0 +1,26 @@
+singularityDir = "/gpfs/scratch/${USER}/singularity_images_nextflow"
+
+//Profile config names for nf-core/configs
+params {
+  config_profile_description = 'Computerome 2.0 cluster profile provided by nf-core/configs.'
+  config_profile_contact = 'Marc Trunjer Kusk Nielsen (@marcmtk)'
+  config_profile_url = 'https://www.computerome.dk/'
+}
+
+singularity {
+  enabled = true
+  autoMounts = true
+  cacheDir = singularityDir
+}
+
+process {
+  beforeScript = 'module load singularity squashfs-tools'
+  executor = 'pbs'
+  queue = { task.memory >= 192.GB ? 'fatnode': 'thinnode'}
+}
+
+params {
+  max_memory = 1500.GB
+  max_cpus = 40
+  max_time = 168.h
+}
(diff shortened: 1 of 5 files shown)


conf/computerome.config
```diff
@@ -0,0 +1,29 @@
+//Profile config names for nf-core/configs
+params {
+  config_profile_description = 'Computerome 2.0 cluster profile provided by nf-core/configs.'
+  config_profile_contact = 'Marc Trunjer Kusk Nielsen (@marcmtk)'
+  config_profile_url = 'https://www.computerome.dk/'
+  project = null
+  schema_ignore_params = "project,genomes,cache_dir,modules"
```
drpatelh, line 7: `genomes` and `modules` should be ignored in the pipeline `nextflow.config` and so don't need to be set here?

drpatelh: Like what you have done with `cache_dir` and `project` though 👌🏽

conf/computerome.config
```diff
@@ -0,0 +1,29 @@
...
+  
+  //Thin nodes with 192GB and Fat nodes with ~1500GB. Torque should be allowed to handle this
+  max_memory = 1500.GB 
+  max_cpus = 40
+  
+  //There is no max walltime on the cluster, but a week seems sensible if not directly specified
+  max_time = 168.h 
+}
+params.cache_dir = "/home/projects/$params.project/scratch"
```
drpatelh, line 16: Specify this in the `params` scope above?

drpatelh (commented):
Looks great overall @marcmtk ! Just a few comments to address. I sent you an invite to become a member of the org too :)

drpatelh (approved):
LGTM! Feel to merge after any tests you need to do.

marcmtk (commented):
Sorry, hadn't seen you'd already approved the changes and cleaned it a little more :)

PR #281, opened 2021-09-21
update sanger profile so that long jobs to be submitted to the long queue

aunderwo:
simple config update to submit taks with `process_long` label to LSF long queue


drpatelh (approved):
Thanks @aunderwo !

PR #286, opened 2021-10-04
Update sarek - cfc specific config

FriederikeHanssen:
The current sarek - cfc config did not respect `--max_memory` etc which is useful in some cases. Adding the `check_resource` function should fix it. Tested if by providing it as custom config.


maxulysse (approved):

ggabernet (approved):
LGTM

Issue #291, opened 2021-10-27
Sarek on Irma breaks with new uppmax.config


parlundin:
When running Sarek on IRMA with updated uppmax.config 

```
Command output:
sbatch: error: Batch job submission failed: Invalid feature specification.
```
`#SBATCH -A ngi2016003  -p node -C mem512GB`

This probably creates issues running other nf-core pipelines on Irma as well.

maxulysse:
ok, so I'm assuming that this PR broke it: https://github.com/nf-core/configs/commit/ac815ce790690754807fbcab1f52444868350d7e

Am I right into assuming that this commit is the last one valid for this config?
https://github.com/nf-core/configs/tree/ff20f121f952a731b93f5719bc2006956117b9c1

Can we download a specific tag of nf-core/configs with nf-core/tools download? like when we use pipelines with `--custom_config_base https://raw.githubusercontent.com/nf-core/configs/ff20f121f952a731b93f5719bc2006956117b9c1`?
That should actually do the trick

ewels:
This comes from the recent merge of #226 from @pontus - it looks like the logic misses Irma, which we just found out about the hard way after our latest deployment of nf-core pipelines.

I think that it basically comes down to this code needing a `<=` instead of `<`:
https://github.com/nf-core/configs/blob/5afa43ad1534d526b4179b20adc598b1de947e63/conf/uppmax.config#L31-L33

Maximum memory jobs  end up requesting exactly 250: https://github.com/nf-core/configs/blob/5afa43ad1534d526b4179b20adc598b1de947e63/conf/uppmax.config#L72-L74

And so they miss that clause and run through to the end, asking for 512 instead.

maxulysse:
I think the underlying issue would be that we would need to be able to download a specific tag/branch/commit whatever with tools, because Sarek 2.7 broke and it was working before.
This is bad for production, we do need to be be able to use fixed config files as well without problem.
I know it's easy when online, but more complicated offline.
I'll raise an issue on nf-core tools for that.

pontus:
:facepalm: 

That's obviously wrong with setting that explicitly as done.

(I'm a bit confused about how your actual issue came about, though, was this a normal increasing retry or do you have custom configuration to set memory to 250 Gbyte?)

pontus:
[...]

PR #293, opened 2021-11-02
Make memory checks use closed intervals

pontus:
Requesting e.g. 250 Gbyte should not require a 512 Gbyte node (or 125 Gbyte a 256 Gbyte).

Addresses part of https://github.com/nf-core/configs/issues/291.


ewels (approved):
Thanks!

PR #294, opened 2021-11-02
UPPMAX profile - minor tweaks

ewels:
* Only one `params` block instead of two
* Return only one `grep` result and don't pipe to `head`
* Use `.startsWith()` instead of regex matching (shouldn't make any difference but is harder to mess up)
* Print detected cluster name in the profile description

Not sure if any of these changes will fix the issues we've seen in #291, but I figure that they're probably all sensible tweaks anyway.

### DIFF
--- a/conf/uppmax.config
+++ b/conf/uppmax.config
@@ -65,7 +65,7 @@ process {
 // Cluster: Snowy
 //   Caution: Bianca nodes will be project name-nodenumber, e.g. sens2021500-001
 //   so cannot rely on just starting with 's'
-if (hostname ==~ "s[0-9][0-9]*") {
+if (hostname.matches("^s[0-9][0-9]*")) {
   params.max_time = 700.h
   params.max_memory = 3880.GB
   params.config_profile_description = 'UPPMAX (Snowy) cluster profile provided by nf-core/configs.'
--- a/conf/uppmax.config
+++ b/conf/uppmax.config
@@ -1,12 +1,14 @@
 // UPPMAX Config Profile
 params {
-  config_profile_description = 'UPPMAX (Rackham/Bianca) cluster profile provided by nf-core/configs.'
+  // Description is overwritten for other clusters below
+  config_profile_description = 'UPPMAX (Bianca) cluster profile provided by nf-core/configs.'
   config_profile_contact = 'Phil Ewels (@ewels)'
   config_profile_url = 'https://www.uppmax.uu.se/'
   project = null
   clusterOptions = null
   schema_ignore_params = "genomes,input_paths,cluster-options,clusterOptions,project"
   save_reference = true
+  // Defaults set for Bianca - other clusters set below
   max_memory = 500.GB
   max_cpus = 16
   max_time = 240.h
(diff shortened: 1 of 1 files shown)


conf/uppmax.config
```diff
@@ -54,26 +60,18 @@ process {
...
-
-  max_memory = 500.GB
-  max_cpus = 16
-  max_time = 240.h
-  // illumina iGenomes reference file paths on UPPMAX
-  igenomes_base = '/sw/data/igenomes/'
-}
-
-if (hostname ==~ "s[0-9][0-9]*") {
+if (hostname.startsWith("s")) {
```
pontus, line 63: This one breaks bianca IIRC (where IIRC nodes will be project name-nodenumber, e.g. sens2021500-001 or similar).

pontus:
Looks good in general.

My current best theory is that the most likely reason for the issues on irma would be `sinfo` not responding properly. Normallly, `sinfo` should be much more likely to work than e.g. `squeue` IIRC, but it obviously can fail if there are issues with slurm. One possibility would be to add a few retries before giving up if it fails.

conf/uppmax.config
```diff
@@ -1,11 +1,17 @@
-// Profile config names for nf-core/configs
+// UPPMAX Config Profile
 params {
-  config_profile_description = 'Swedish UPPMAX cluster profile provided by nf-core/configs.'
+  config_profile_description = 'UPPMAX (Rackham) cluster profile provided by nf-core/configs.'
```
mahesh-panchal, line 3:
```suggestion
  config_profile_description = 'UPPMAX (Rackham/Bianca) cluster profile provided by nf-core/configs.'
```

conf/uppmax.config
```diff
@@ -54,26 +60,18 @@ process {
...
 }
 
-if (hostname ==~ "i.*") {
+if (hostname.startsWith("i")) {
   params.max_memory = 250.GB
+  params.config_profile_description = 'UPPMAX (Irma) cluster profile provided by nf-core/configs.'
 }
 
-if (hostname ==~ "r.*") {
+if (hostname.startsWith("r")) {
```
mahesh-panchal, line 74: Why don't we use these as the defaults above if Rackham is the default machine?

conf/uppmax.config
maxulysse, line 3: This is not only Rackham since it works on Irma and Bianca too

mahesh-panchal: Lower down, Irma and Snowy are accounted for.

ewels: I'll add a code comment noting that

conf/uppmax.config
ewels: Yeah good question. I'll update it to do that instead.

ewels: ah maybe because the default is Rackham / Bianca and not just Rackham.

mahesh-panchal: Then we should add a comment above, that those are for Bianca, while these are for Rackham, and split the descriptions too.

ewels:
Thanks @pontus / @mahesh-panchal / @maxulysse - comments all now addressed I think. Shout if I missed anything.

@pontus maybe you can sanity check that the top-level max memory / cpus are sane defaults for Bianca and that my code comments are all correct? Thanks!

conf/uppmax.config
```diff
@@ -60,22 +62,32 @@ process {
   scratch = '$SNIC_TMP'
 }
 
-if (hostname.startsWith("s")) {
+// Cluster: Snowy
+//   Caution: Bianca nodes will be project name-nodenumber, e.g. sens2021500-001
+//   so cannot rely on just starting with 's'
+if (hostname ==~ "s[0-9][0-9]*") {
```
mahesh-panchal, line 68:
```suggestion
if (hostname.matches("^s[0-9][0-9]*")) {
```
Maybe use this if you don't want the `==~`

pontus (approved):
Looks good as far as I can see.

mahesh-panchal (approved):

maxulysse (approved):
<3

PR #295, opened 2021-11-02
Change params.hostnames from map to string

drpatelh:
See https://github.com/nf-core/tools/issues/1304

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,12 +66,6 @@ See [`nf-core/configs`](https://github.com/nf-core/configs/tree/master/conf) for
 Please also make sure to add an extra `params` section with `params.config_profile_description`, `params.config_profile_contact` and `params.config_profile_url` set to reasonable values.
 Users will get information on who wrote the configuration profile then when executing a nf-core pipeline and can report back if there are things missing for example.
 
-### Checking user hostnames
-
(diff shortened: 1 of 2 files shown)


drpatelh:
Tested with the command below to pull in and use the branch in this PR and everything is working as expected:

```
NXF_VER=21.10.0-SNAPSHOT \
nextflow run main.nf \
    -profile test,singularity \
    --custom_config_base 'https://raw.githubusercontent.com/drpatelh/nf-core-configs/hostnames'
```

We can remove `params.hostnames` entirely from this repo because it is already initialised in (`nextflow.config`)[https://github.com/nf-core/rnaseq/blob/964425e3fd8bfc3dc7bce43279a98d17a874d3f7/nextflow.config#L106]. That will evaluate to `false` because it is an empty Map and hence bypass the Schema validation code so we can just dump it entirely.

This will fix the NF compatibility issue outlined in https://github.com/nf-core/tools/issues/1304 within the context of nf-core pipelines. The next and more longer term fixes are to remove `params.hostnames` in the nf-core pipeline template and to change the `type` from a string to a map where relevant in the Schema.

ewels (approved):

PR #300, opened 2021-12-03
eager @ eva: reintroduce bionode8, bump overhead for flagstat

jfy133:
Now found a log with sufficient information, it was a memory exeedence (or insufficient memory on the node once assigned) issue.


TCLamnidis (approved):
LGTM

PR #313, opened 2022-01-06
Added memory & cluster options specification for eddie sarek config

ameynert:
Running Sarek 2.7.1 with -profile test,eddie on University of Edinburgh cluster was failing on MapReads. Fixed with additional specification of memory and clusterOptions as per other processes.


pontus (approved):
See no issue.

PR #320, opened 2022-02-02
Few changes to CBE module loading

rbpisupati:
There have been changes to module loading in CBE!

Also turned off singularity by default. It is creating error while running conda environments.


phue:
Hi Rahul,

I'm not in favour of removing singularity as the default. Using containers is the preferred way to run nf-core pipelines and that is what most cbe users are doing.

Loading the conda module in the profile was meant as a fallback option and if it causes problems I would rather get rid of conda to be honest. What is the error you are seeing exactly?

rbpisupati:
I made this sometime back and dont remember the error. 

Ok, can you just push changes related to conda module?

phue (approved):

phue:
Alright, thanks for bringing this up.

While I was here I tweaked the queue selection and added initial support for the gpu queue

PR #322, opened 2022-02-04
Add UAB Cheaha profile

lianov:
Request to add profile for the UAB HPC cluster, Cheaha.

All required changes are present in this request (`cheaha.md` added to `docs`, `cheaha.conf` added to `conf`, profile added to `nfcore_custom.config` and profile added to `README.md`).

The configuration has been tested with the rnaseq and nanoseq pipelines.

Contributors to profile, review and tests are Austyn Trull (@atrull314) and Lara Ianov (@lianov).

### DIFF
--- a/docs/cheaha.md
+++ b/docs/cheaha.md
@@ -16,4 +16,17 @@ module load Nextflow
 All of the intermediate files required to run the pipeline will be stored in the `work/` directory. It is recommended to delete this directory after the pipeline has finished successfully because it can get quite large, and all of the main output files will be saved in the `results/` directory anyway.
 
 >NB: You will need an account to use the HPC cluster on Cheaha in order to run the pipeline. If in doubt contact UAB IT Research Computing.
->NB: Nextflow will need to submit the jobs via SLURM to the HPC cluster and as such the commands above will have to be executed on one of the login nodes. If in doubt contact UAB IT Research Computing.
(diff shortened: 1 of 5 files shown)


docs/cheaha.md
```diff
@@ -0,0 +1,32 @@
...
+Before running the pipeline you will need to load Singularity and Nextflow using the environment module system on Cheaha. You can do this by issuing the commands below:
+
+```bash
+## Singularity environment modules
+module purge
+module load Singularity
+module load Nextflow
+```
+
+All of the intermediate files required to run the pipeline will be stored in the `work/` directory. It is recommended to delete this directory after the pipeline has finished successfully because it can get quite large, and all of the main output files will be saved in the `results/` directory anyway.
```
jfy133, line 16: FYI: you could also consider adding `cleanup = true` to the profile to have this to auto clean up.

jfy133 (approved):

docs/cheaha.md
lianov: Thanks for the suggestion @jfy133 . I think we will keep this option off since we find some components of the pipeline in the `work` directory useful for testing pipelines and debugging when needed, and others may benefit from it too without auto-deletion. So a more curated approach to this directory may be overall more beneficial.

PR #323, opened 2022-02-04
Added ARTIC v4.1 primer set information

stevin-wilson:
ARTIC v4.1 has been recommended to improve sequencing performance on Omicron samples [1], and including information regarding the v4.1 primer set (obtained from [2]) will be convenient for the users of nf-core/viralrecon. Thank you.

[1] https://www.ecdc.europa.eu/sites/default/files/documents/Methods-for-the-detection-and-characterisation-of-SARS-CoV-2-variants-first-update.pdf
[2] https://github.com/artic-network/artic-ncov2019/tree/master/primer_schemes/nCoV-2019/V4.1


drpatelh (approved):
Thanks @stevin-wilson !

PR #324, opened 2022-02-06
New attempt of workaround SGE + Java incompatibility for eager @ EVA

jfy133:


conf/pipeline/eager/eva.config
```diff
@@ -3,7 +3,7 @@
 params {
   // Specific nf-core/configs params
   config_profile_contact = 'James Fellows Yates (@jfy133)'
-  config_profile_description = 'nf-core/eager EVA profile provided by nf-core/configs'
+  config_profile_description = 'nf-core/eager EVA profile provided by nf-core/configs DEV'
```
TCLamnidis, line 6: Should this be DEV?

TCLamnidis (approved):
LGTM, except for the DEV in the description.
Are all these codes really needed in all the tools?

conf/pipeline/eager/eva.config
jfy133: Not any more!

jfy133:
Yes... don't want ot put the codes in non-Java tools as those actually work correctly... unfotunately we have a lot of java tools...

conf/pipeline/eager/eva.config
jfy133, line 6:
```suggestion
  config_profile_description = 'nf-core/eager EVA profile provided by nf-core/configs'
```

TCLamnidis:
What I meant was do all the Java tools need all these exit codes, or did you apply them all across the board for consistency?

TCLamnidis:
Also, if you have a list of what each code you added means to append here, I would not be against it :eyes:

jfy133:
> What I meant was do all the Java tools need all these exit codes, or did you apply them all across the board for consistency?

Consistency, but all Java ones need the exit code 1, as all their errors results in that. I kept the other things in tehe as I couldn't remeber if they are inherited or overwritten from the base exit codes in the pipeline.

> Also, if you have a list of what each code you added means to append here, I would not be against it eyes

Varies per tool slightly, they just _generally_ mean the same thing...ish...

PR #337, opened 2022-03-02
Update mpcdf.config

maltesemike:
Update max_memory to 120.GB (in MB) to match the resources allowed on the raven cluster.


ilight1542:
I  think we would rather want to change the max resources to the actual maximum available (2TB and 72 CPUS), and then deal with the shared compute nodes via withProcess flags @jfy133, thoughts?

jfy133:
Ha, yes: I half-finished a review saying the same thing

image

jfy133:
https://github.com/nf-core/configs/pull/339

PR #338, opened 2022-03-02
Update mpcdf.md

maltesemike:
Updates resources to match the changes in the config file for raven. These have been confirmed to work for shared users. Also changed the order in which the -profile flag should be used to match that given in the readme above.


jfy133:
Closing for https://github.com/nf-core/configs/pull/339

PR #340, opened 2022-03-03
modified mpcdf configs (general + eager)

ilight1542:
changed config file for mpcdf raven (general to be 2048 GB max mem, and 72 CPUS max) and eager with malt using the full large node and the mc_huge label changing to a full node, of comprable size to avoid shared node allocation.


ilight1542:
will merge with RNAseq changes @jfy133

PR #346, opened 2022-03-12
Added SAHMRI profile

nathanhaigh:


nathanhaigh:
@jfy133 - How do I get this PR reviewed?

conf/sahmri.config
```diff
@@ -0,0 +1,29 @@
...
+  queueSize       = 50
+  submitRateLimit = '10 sec'
+}
+singularity {
+  enabled         = true
+  autoMounts      = true
+  beforeScript    = 'export PATH=/apps/opt/singularity/latest/bin:${PATH}'
+  cacheDir        = '/cancer/storage/shared/simg'
+}
+cleanup = true
```
jfy133, line 29: If I set this to run by default within the config, I normally like to have a `debug` profile or something to allow a user to turn this off (e.g. to allow checking of `work/` dirs while developing a pipeline)

conf/sahmri.config
```diff
@@ -0,0 +1,29 @@
...
+  igenomes_base   = '/cancer/storage/shared/igenomes/'
+}
+process {
+  executor        = 'slurm'
+  queue           = 'sahmri_prod_hpc,sahmri_cancer_hpc'
+  maxRetries      = 2
+
+  cpus            = { check_max( 2 * task.attempt, 'cpus') }
+  memory          = { check_max( 1.GB * task.attempt, 'memory') }
+  time            = { check_max( 10.m * task.attempt, 'time') }
```
jfy133, line 17: Is there a purpose for setting this here? All nf-ore pipelines define default settings for these on a pipeline-basis so shouldn't be necessary. Or is the purpose here to support non-nf-core pipelines as well?

docs/sahmri.md
```diff
@@ -0,0 +1,18 @@
...
+To use, run the pipeline with `-profile sahmri`. This will download and launch the [`sahmri.config`](../conf/sahmri.config) which has been pre-configured
+with a setup suitable for the SAHMRI HPC cluster. Using this profile, either a docker image containing all of the required software will be downloaded,
+and converted to a Singularity image or a Singularity image downloaded directly before execution of the pipeline.  
+
+The latest version of Nextflow is not installed by default on the SAHMRI HPC cluster. You will need to install it into a directory you have write access to.
+Follow these instructions from the Nextflow documentation.
+
+- Install Nextflow : [here](https://www.nextflow.io/docs/latest/getstarted.html#)
+
+All of the intermediate files required to run the pipeline will be stored in the `work/` directory. It is recommended to delete this directory after the pipeline
```
jfy133, line 13: You've set `cleanup` to true, so this will happen automatically. I would mabe clarify that this therefore is not actually necessary unless you first have multiple failed runs and had to resume.

jfy133 (changes requested):
Please also update README.md to include `sahmri` in the list, but other than my questions (rather than issues), LGTM.

apeltzer (approved):

conf/sahmri.config
nathanhaigh: The intention is to also use this for non nf-core pipelines too.

jfy133 (approved):

PR #354, opened 2022-03-17
Use prettier

Emiller88:


.github/workflows/linting.yml
```diff
@@ -1,18 +1,22 @@
-name: Markdown linting
-# This workflow is triggered on pushes and PRs to the repository.
-on: [push, pull_request]  
+name: Code Linting
+on:
+  push:
```
ewels, line 3: Why only master?

ewels (approved):

.github/workflows/linting.yml
Emiller88: Probably related to https://github.com/nf-core/nf-co.re/issues/1105

Should be fixed in 42bb08dab43240e92ce5da54c23264e38eed1d1a and we can address it if that didn't work

Emiller88:
I think someone needs to update the require statuses 
image

Let's get that addressed before merging.

ewels:
Done ✅

ewels (approved):
LGTM!

PR #357, opened 2022-03-25
Execute sinfo with bash shell

mahesh-panchal:
Fixes the piped command execution

### DIFF
--- a/conf/uppmax.config
+++ b/conf/uppmax.config
@@ -23,7 +23,7 @@ singularity {
 
 def hostname = "r1"
 try {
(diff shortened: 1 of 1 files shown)


pontus (approved):
I don't know (or can't read from docs and understand) why the old version stopped working and this works, but if it does that's fine.

mahesh-panchal:
It seems to have been something for a long time. 
I found this from 2007. 
http://devilelephant.blogspot.com/2007/03/groovy-process-piping.html

PR #389, opened 2022-06-08
[WIP] Add custom configuration for Sage Bionetworks 

BrunoGrandePhD:
---
name: `sage.config`
about: A custom Nextflow configuration for Sage Bionetworks
---

- [x] (N/A) You've included links to relevant issues, if any

### DIFF
--- a/conf/sage.config
+++ b/conf/sage.config
@@ -38,7 +38,7 @@ process {
   }
 
   // Preventing Sarek labels from using the actual maximums
-  withLabel:max_memory {
+  withLabel:memory_max {
     memory = { check_max( 128.GB * task.attempt,   'memory' ) }
   }
   withLabel:cpus_max {
--- a/conf/sage.config
+++ b/conf/sage.config
@@ -1,0 +1,58 @@
+params {
+  config_profile_description = 'The Sage Bionetworks profile'
+  config_profile_contact = 'Bruno Grande (@BrunoGrandePhD)'
+  config_profile_url = 'https://github.com/Sage-Bionetworks-Workflows'
+}
+
+process {
+
+  cpus   = { check_max( 1    * slow(task.attempt), 'cpus'   ) }
+  memory = { check_max( 6.GB * task.attempt,       'memory' ) }
+  time   = { check_max( 24.h * task.attempt,       'time'   ) }
+
+  errorStrategy = { task.exitStatus in [143,137,104,134,139,247] ? 'retry' : 'finish' }
+  maxRetries    = 5
+  maxErrors     = '-1'
+
+  // Process-specific resource requirements
+  withLabel:process_low {
+    cpus   = { check_max( 4     * slow(task.attempt),  'cpus'   ) }
+    memory = { check_max( 12.GB * task.attempt,        'memory' ) }
+    time   = { check_max( 24.h  * task.attempt,        'time'   ) }
+  }
+  withLabel:process_medium {
+    cpus   = { check_max( 12    * slow(task.attempt), 'cpus'   ) }
+    memory = { check_max( 36.GB * task.attempt,       'memory' ) }
+    time   = { check_max( 48.h  * task.attempt,       'time'   ) }
+  }
+  withLabel:process_high {
+    cpus   = { check_max( 24    * slow(task.attempt), 'cpus'   ) }
+    memory = { check_max( 72.GB * task.attempt,       'memory' ) }
+    time   = { check_max( 96.h  * task.attempt,       'time'   ) }
+  }
+  withLabel:process_long {
+    time   = { check_max( 192.h  * task.attempt,   'time'   ) }
+  }
+  withLabel:process_high_memory {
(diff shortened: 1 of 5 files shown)


thomasyu888:
LGTM

thomasyu888:
LGTM

conf/sage.config
```diff
@@ -0,0 +1,58 @@
...
+
+  cpus   = { check_max( 1    * slow(task.attempt), 'cpus'   ) }
+  memory = { check_max( 6.GB * task.attempt,       'memory' ) }
+  time   = { check_max( 24.h * task.attempt,       'time'   ) }
+
+  errorStrategy = { task.exitStatus in [143,137,104,134,139,247] ? 'retry' : 'finish' }
+  maxRetries    = 5
+  maxErrors     = '-1'
+
+  // Process-specific resource requirements
```
wpoehlm, line 17: I'm not sure that there's necessarily a positive correlation between memory requirements and expected runtime (for example tools/scripts that use only one cpu and only a small amount of RAM but run for a long time), but I don't have a problem with this since these are always customizable by the user

wpoehlm (approved):
Looks good to me, it will be nice to use the new sage-igenomes path more easily!

conf/sage.config
```diff
@@ -0,0 +1,58 @@
...
+    cpus   = { check_max( 24    * slow(task.attempt), 'cpus'   ) }
+    memory = { check_max( 72.GB * task.attempt,       'memory' ) }
+    time   = { check_max( 96.h  * task.attempt,       'time'   ) }
+  }
+  withLabel:process_long {
+    time   = { check_max( 192.h  * task.attempt,   'time'   ) }
+  }
+  withLabel:process_high_memory {
+    memory = { check_max( 200.GB * task.attempt,   'memory' ) }
+  }
```
wpoehlm, line 38: I'm not sure that there's necessarily a positive correlation between memory requirements and expected runtime (for example tools/scripts that use only one cpu and only a small amount of RAM but run for a long time), but I don't have a problem with this since these are always customizable by the user

conf/sage.config
```diff
@@ -0,0 +1,58 @@
...
+
+}
```
wpoehlm, line 40: I'm not sure that there's necessarily a positive correlation between memory requirements and expected runtime (for example tools/scripts that use only one cpu and only a small amount of RAM but run for a long time), but I don't have a problem with this since these are always customizable by the user

conf/sage.config
```diff
@@ -0,0 +1,58 @@
...
+  max_cpus      = 64
+  max_time      = 168.h  // One week
+}
+
+// Function to slow the increase of the resource multipler
+// as attempts are made. The rationale is that some CPUs
+// don't need to be increased as fast as memory.
+def slow(attempt, factor = 2) {
+  return Math.ceil( attempt / factor) as int
+}
```
maxulysse, line 58: I like this level of customization, but why not directly changing
```
cpus   = { check_max( 4     * slow(task.attempt),  'cpus'   ) }
```
to
```
cpus   = { check_max( 2     * task.attempt,  'cpus'   ) }
```

conf/sage.config
BrunoGrandePhD: I totally agree, but this is the general model that nf-core uses ([example](https://github.com/nf-core/rnaseq/blob/master/conf/base.config#L21-L42)). They label processes with one of these to generally capture the "resource intensity", but I would like to consider memory- and CPU-intensive processes separately so we can optimize resource allocation. 

In general, I've been finding that our workflows aren't using all of the CPUs at their disposal, so I increased the CPU-to-memory ratio for these general labels. I also plan to create pipeline-specific config files once I delve into the resource stats of actual workflow runs, which I can add to the `sage` profile.

conf/sage.config
BrunoGrandePhD: In this particular case, I don't want to decrease the initial number of allocated CPU cores. For example:

```
cpus   = { check_max( 4     * slow(task.attempt),  'cpus'   ) }
```

would result in these CPU requests over five attempts: 4, 4, 8, 8, and 12. 

Whereas:

```
cpus   = { check_max( 2     * task.attempt,  'cpus'   ) }
```

would result in these CPU requests over five attempts: 2, 2, 4, 4, and 6. 

The former "progression" matches what I would like to see. In other words, try increasing the memory first without changing the number of CPU cores (since memory is more commonly the limiting factor), and if you need to re-run a third time, then try increasing the number of CPU cores.

ewels:
You do not need to include the `check_max()` function in configs hosted in nf-core/configs, no. These are included into `nextflow.config` by the run and it's present there already.

The time you _do_ need to copy this function into a config file is if you're doing `nextflow run -c myconfig.conf`. Then `myconfig.conf` is parsed separately to the pipeline config and so you need the function in the custom config file.

Does that make sense?

ewels (approved):
Failing test is an external issue, PR looks good to me 👍🏻

PR #395, opened 2022-07-01
New config: Add config for Medair

Gwennid:
----
name: medair.config
about: A custom Nextflow configuration for Clinical Genomics Gothenburg
----

This PR is to add a config for the medair cluster, which is the cluster used at Clinical Genomics Gothenburg, Sweden (https://www.scilifelab.se/units/clinical-genomics-goteborg/).

### DIFF
--- a/conf/medair.config
+++ b/conf/medair.config
@@ -1,13 +1,13 @@
 //Profile config names for nf-core/configs
 params {
   config_profile_description = 'Cluster profile for medair (local cluster of Clinical Genomics Gothenburg)'
-  config_profile_contact = 'Clinical Genomics, Gothenburg'
-  config_profile_url = 'https://www.scilifelab.se/units/clinical-genomics-goteborg/'
+  config_profile_contact     = 'Clinical Genomics, Gothenburg'
+  config_profile_url         = 'https://www.scilifelab.se/units/clinical-genomics-goteborg/'
 }
 
 //Nextflow parameters
 singularity {
-    enabled = true
+    enabled  = true
     cacheDir = "/apps/bio/dependencies/nf-core/singularities"
 }
 
@@ -15,26 +15,32 @@ profiles {
   
   wgs {
     process {
-      queue = 'wgs.q'
-      executor = 'sge'
-      penv = 'mpi'
+      queue                  = 'wgs.q'
+      executor               = 'sge'
+      penv                   = 'mpi'
       process.clusterOptions = '-l excl=1'
-      params.max_cpus = 40
-      params.max_time = 48.h
-      params.max_memory = 128.GB
+      params.max_cpus        = 40
+      params.max_time        = 48.h
+      params.max_memory      = 128.GB
     }
   }
 
   production {
     process {
-      queue = 'production.q'
-      executor = 'sge'
-      penv = 'mpi'
+      queue                  = 'production.q'
+      executor               = 'sge'
+      penv                   = 'mpi'
       process.clusterOptions = '-l excl=1'
-      params.max_cpus = 40
-      params.max_time = 480.h
-      params.max_memory = 128.GB
+      params.max_cpus        = 40
+      params.max_time        = 480.h
+      params.max_memory      = 128.GB
     }
   }
 }
 
+//Specific parameter for pipelines that can use Sentieon (e.g. nf-core/sarek, nf-core/raredisease)
+process {
+  withLabel:'sentieon' {
+    container = "/apps/bio/singularities/sentieon-211204-peta.simg"
+  }
+}
--- a/conf/medair.config
+++ b/conf/medair.config
@@ -1,0 +1,39 @@
+//Profile config names for nf-core/configs
+params {
+  config_profile_description = 'Cluster profile for medair (local cluster of Clinical Genomics Gothenburg)'
+  config_profile_contact = 'Clinical Genomics, Gothenburg'
+  config_profile_url = 'https://www.scilifelab.se/units/clinical-genomics-goteborg/'
+}
+
+//Nextflow parameters
+singularity {
+    enabled = true
+}
+
+profiles {
+  
+  wgs {
+    process {
+      queue = 'wgs.q'
+      executor = 'sge'
+      penv = 'mpi'
+      process.clusterOptions = '-l excl=1'
+      params.max_cpus = 40
+      params.max_time = 48.h
+      params.max_memory = 128.GB
+    }
+  }
+
+  production {
+    process {
+      queue = 'production.q'
+      executor = 'sge'
+      penv = 'mpi'
+      process.clusterOptions = '-l excl=1'
+      params.max_cpus = 40
+      params.max_time = 480.h
+      params.max_memory = 128.GB
+    }
+  }
+}
+
--- a/conf/medair.config
+++ b/conf/medair.config
@@ -1,7 +1,7 @@
 //Profile config names for nf-core/configs
 params {
   config_profile_description = 'Cluster profile for medair (local cluster of Clinical Genomics Gothenburg)'
-  config_profile_contact     = 'Clinical Genomics, Gothenburg'
+  config_profile_contact     = 'Clinical Genomics, Gothenburg (cgg-rd@gu.se, cgg-it@gu.se)'
   config_profile_url         = 'https://www.scilifelab.se/units/clinical-genomics-goteborg/'
 }
 
--- a/conf/medair.config
+++ b/conf/medair.config
@@ -8,6 +8,7 @@ params {
 //Nextflow parameters
 singularity {
     enabled = true
+    cacheDir = "/apps/bio/dependencies/nf-core/singularities"
 }
 
 profiles {
--- a/docs/medair.md
+++ b/docs/medair.md
@@ -57,7 +57,7 @@ module load singularity
 
 ### Choose a profile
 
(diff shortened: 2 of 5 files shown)


docs/medair.md
```diff
@@ -0,0 +1,70 @@
...
+Use the `nf-core download --singularity-cache-only` command to start a download. It will open an interactive menu. Choose `singularity` for the software container image, and `none` for the compression type.
+
+### Storage of Singularity images
+
+When downloading a new nf-core pipeline for the first time (or a specific version of a pipeline), you can choose to store the Singularity image for future use. A central location for these images is: `/apps/bio/dependencies/nf-core/singularities`
+
+Cached Singularity images can be accessed by running (or adding to your `.bashrc`) the following: 
+
+```
+export NXF_SINGULARITY_CACHEDIR="/apps/bio/dependencies/nf-core/singularities"
```
pbiology, line 36: Could this somehow be be included in the profile? That way we wouldn't have to set it in the users shell

docs/medair.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Set-up
+
+Before running a pipeline you will need to load Nextflow and Singularity using the environment module system on medair. You can do this by issuing the commands below:
+
+```bash
+## Load Nextflow and Singularity environment modules
+module purge
+module load nextflow/21.10.5.5658
+module load singularity/v3.4.0
```
pbiology, line 51: Perhaps we shouldn't specify versions here. Just `module load nextflow` should load latest version. Same with singularity

docs/medair.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+>Should we link to Confluence page about new modules? See something about updating modules? etc.
```
pbiology, line 54: Don't think we should link to internal documentation in here?

docs/medair.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Choose a profile
+
+Depending on what you are running, you can choose between the `wgs` and `production` profiles. Jobs running with the `wgs` profile run on a queue with higher priority. Jobs running with the `production` profile can last longer (max time: 20 times, versus 2 days for the `wgs` profile).
+
+>Usage: -profile medair,wgs ?? (Check)
+
+## iGenomes specific configuration
+
+>TODO modify this part. Do we want to have the iGenomes somewhere?
```
pbiology, line 64: I have some of them downloaded. But i think the whole iGenomes thing is being replaced throughout all of nf-core, so don't think we should include it here

pbiology: And isn't this more of a pipeline specific thing anyway?

docs/medair.md
```diff
@@ -0,0 +1,70 @@
...
+
+A local copy of the iGenomes resource has been made available on PROFILE CLUSTER so you should be able to run the pipeline against any reference available in the `igenomes.config` specific to the nf-core pipeline.
+You can do this by simply using the `--genome <GENOME_ID>` parameter.
+
+>NB: You will need an account to use the HPC cluster on PROFILE CLUSTER in order to run the pipeline. If in doubt contact IT.
+>NB: Nextflow will need to submit the jobs via the job scheduler to the HPC cluster and as such the commands above will have to be executed on one of the login nodes. If in doubt contact IT.
```
EmmaCAndersson, line 70: Should it be more specific than IT. Like cgg-it?

pbiology: yeah, good point

docs/medair.md
Gwennid: Sounds reasonable

docs/medair.md
Gwennid: At least the pipelines I ran can use iGenomes, so I do not know how pipeline specific it is or not. But if it's going to be replaced we might as well remove it!

docs/medair.md
Gwennid: I mentioned cgg-it instead of "IT" and I changed the location of where cgg-it is mentioned, to locations that made more sense.

conf/medair.config
```diff
@@ -0,0 +1,39 @@
...
+      executor = 'sge'
+      penv = 'mpi'
+      process.clusterOptions = '-l excl=1'
+      params.max_cpus = 40
+      params.max_time = 480.h
+      params.max_memory = 128.GB
+    }
+  }
+}
+
```
sima-r, line 39: Är det inte värt att ha path till sentieon container med i denna config filen?  
process {
    withLabel:'sentieon' {
        container = "/apps/bio/singularities/sentieon-211204-peta.simg"
    }
}

sima-r (approved):
Looks good. I  had a suggestion and have no more comments.

docs/medair.md
Gwennid: Yes apparently it can  be included. I did it now. We should verify that it works though (I will write a comment if I test it).

conf/medair.config
Gwennid: That's a good idea. I didn't find any example of another cluster config that does it. But in the sarek specific config for the munin cluster it was done: `conf/pipeline/sarek/munin.config`
What do we think? Should that go to the general config or to a nf-core/raredisease specific one?

sima-r: I'd actually like to have it in the cluster config file which I assume we will use for the most of Nextflow pipelines and also test runs. So I think it's better to already have it here than specifying it in other configs where we probably want to specify some input/output paths and other variables. (But I know that not in all of the pipelines we run sentieon )

Gwennid: Hum - what I meant is that one can have a cluster & pipeline specific config, that is also stored in nf-core/configs. But I agree, it does not harm to have it there even though not all pipelines use Sentieon.
What about the Sentieon version? Should we go with the more recent version as you suggest, or start with the version used in WOPR and change once we have done the comparisons?

docs/medair.md
Gwennid: I tried to test it and in fact I do not know how to...
The cache directory is used e.g. for `nf-core download`, but that is not a Nextflow command and we do not give our cluster config there...
@sima-r , @pbiology , any idea?

pbiology: But it's also to tell the pipe where to look for singularities, no?

Gwennid: Yes, that part is fine (running pipelines).
But when one downloads a new pipeline and wants to store the image, the cluster config is not involved. So I don't see how to export the variable in bashrc or such. Or am I missing something totally?

conf/medair.config
sima-r: Another question, do we want to call it medair?

EmmaCAndersson (approved):

conf/medair.config
pbiology: I think medair is a suitable name. It's overall what we call the cluster

pbiology (approved):

conf/medair.config
```diff
@@ -0,0 +1,46 @@
+//Profile config names for nf-core/configs
+params {
+  config_profile_description = 'Cluster profile for medair (local cluster of Clinical Genomics Gothenburg)'
+  config_profile_contact     = 'Clinical Genomics, Gothenburg'
```
jfy133, line 4: It would be good to have this a bit more specific. Either a specific person who runs nf-core pipelines regularly, or if you mean e.g. IT department (if they are maintaining the profile)

docs/medair.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+### Storage of Singularity images
+
+When downloading a nf-core pipeline for the first time (or a specific version of a pipeline), you can choose to store the Singularity image for future use. A central location for these images is: `/apps/bio/dependencies/nf-core/singularities`.
+
+If you run or add the following to your `.bashrc`, Nexflow will know where to store the images:
+
+```bash
+export NXF_SINGULARITY_CACHEDIR="/apps/bio/dependencies/nf-core/singularities"
```
jfy133, line 32: This is a little confusing, by your config above says it'll always go to the `singularities` folder? A user doesn't ave an option in this case.

docs/medair.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Set-up: load Nextflow and Singularity
+
+Before running a pipeline you will need to load Nextflow and Singularity using the environment module system on Medair. You can do this by issuing the commands below:
+
+```bash
+## Load Nextflow and Singularity environment modules
+module purge
+module load nextflow
+module load singularity
```
jfy133, line 55: Repetition of lines 15-22?myabe you can just refer back to that?

jfy133 (changes requested):
The configs themselves look good to me, just some comments on the docs

docs/medair.md
Gwennid: True. I clarified it now.

conf/medair.config
mashehu: Agree. [Clinical Genomics Stockholm](https://github.com/nf-core/configs/blob/master/conf/hasta.config) did the same... It is in the tutorial, but maybe we should also clarify it in the repo readme?

docs/medair.md
Gwennid: I agree it looks repetitive, but it's not exactly the same things we load (nextflow + conda environment to download pipelines, nextflow + singularity to run) and we thought that if one only wants to run a pipeline, one might jump directly to that section...

jfy133: Ok fair enough!

conf/medair.config
jfy133: Hm, I guess that should be updated...

https://github.com/nf-core/nf-co.re/pull/1225 for improving this in the tutorial at least.

jfy133 (approved):

PR #404, opened 2022-08-24
Cheaha Config Update

atrull314:
---
name: Update to existing cheaha.config
about: Merging contributions made by Brandon Wilk (@wilkb777). These changes modify the Cheaha profile so that singularity containers use $USER_SCRATCH
---

### DIFF
--- a/conf/cheaha.config
+++ b/conf/cheaha.config
@@ -5,9 +5,15 @@ params {
     config_profile_url = 'https://www.uab.edu/cores/ircp/bds'
 }
 
+env {
+    TMPDIR="$USER_SCRATCH"
+    SINGULARITY_TMPDIR="$USER_SCRATCH"
+}
+
 singularity {
     enabled = true
     autoMounts = true
+    runOptions = "--contain --workdir $USER_SCRATCH"
 }
 
 process {
--- a/docs/cheaha.md
+++ b/docs/cheaha.md
@@ -13,6 +13,8 @@ module load Singularity
 module load Nextflow
 ```
 
+Various tasks will be run inside of Singularity containers and all temp files typically written to `/tmp` and `/var/tmp` are instead written to the path pointed to by the `USER_SCRATCH` environment variable. This means that these temp files are stored in a user specific location, making them inaccessible to other users for pipeline reruns. Some of these temp files can be large and cleanup is also the responsibility of the user.
+
 All of the intermediate files required to run the pipeline will be stored in the `work/` directory. It is recommended to delete this directory after the pipeline has finished successfully because it can get quite large, and all of the main output files will be saved in the `results/` directory anyway.
 
 > NB: You will need an account to use the HPC cluster on Cheaha in order to run the pipeline. If in doubt contact UAB IT Research Computing.</br></br>


jfy133 (commented):
LGTM but make sure to update the docs so a user knows that singularity images will be saved in `$USER_SCRATCH`

wilkb777:
> make sure to update the docs so a user knows that singularity images will be saved in `$USER_SCRATCH`

@jfy133 thank you for the quick review! I just want to make sure I'm not missing something here or did something wrong, I don't understand how this changes where singularity images are stored?

The addition of the environment variables was following advice in [this issue](https://github.com/nextflow-io/nextflow/issues/2685) to resolve singularity from using `/tmp` on device for containers for executing tasks. I didn't think it would affect the default location where Nextflow caches images.

The goal of the change was to prevent running Singularity containers from thinking they should mount `/tmp` on the device inside the container since the actual temp space on device is very small and the use of `$USER_SCRATCH` is encouraged highly for all temp file needs on Cheaha.

Perhaps I missing something with the `runOptions = "--contain --workdir $USER_SCRATCH"` specification? From the docs this is the listed way to have runtime options passed to singularity and the options were added so that Singularity would automatically mount temp locations inside the container to the scratch space according to Singularity's docs:

```
--contain                use minimal /dev and empty other directories (e.g. /tmp and $HOME) instead of sharing filesystems from your host
--workdir string         working directory to be used for /tmp, /var/tmp and $HOME (if -c/--contain was also used)
```

Granted I also use the `NXF_SINGULARITY_CACHEDIR` environment variable and the `singularity.cacheDir` setting in the Nextflow config file in a lab specific manner that doesn't fit a general config like this and perhaps that's why I hadn't noticed this affecting image caching.

Happy to add updated docs and noting that the `NXF_SINGULARITY_CACHEDIR` and the `singularity.cacheDir` setting should be used to prevent this if that's the case.

jfy133:
> > make sure to update the docs so a user knows that singularity images will be saved in `$USER_SCRATCH`
> 
> @jfy133 thank you for the quick review! I just want to make sure I'm not missing something here or did something wrong, I don't understand how this changes where singularity images are stored?
> 
> The addition of the environment variables was following advice in [this issue](https://github.com/nextflow-io/nextflow/issues/2685) to resolve singularity from using `/tmp` on device for containers for executing tasks. I didn't think it would affect the default location where Nextflow caches images.
> 
> The goal of the change was to prevent running Singularity containers from thinking they should mount `/tmp` on the device inside the container since the actual temp space on device is very small and the use of `$USER_SCRATCH` is encouraged highly for all temp file needs on Cheaha.
> 
> Perhaps I missing something with the `runOptions = "--contain --workdir $USER_SCRATCH"` specification? From the docs this is the listed way to have runtime options passed to singularity and the options were added so that Singularity would automatically mount temp locations inside the container to the scratch space according to Singularity's docs:
> 
> ```
> --contain                use minimal /dev and empty other directories (e.g. /tmp and $HOME) instead of sharing filesystems from your host
> --workdir string         working directory to be used for /tmp, /var/tmp and $HOME (if -c/--contain was also used)
> ```
> 
> Granted I also use the `NXF_SINGULARITY_CACHEDIR` environment variable and the `singularity.cacheDir` setting in the Nextflow config file in a lab specific manner that doesn't fit a general config like this and perhaps that's why I hadn't noticed this affecting image caching.
> 
> Happy to add updated docs and noting that the `NXF_SINGULARITY_CACHEDIR` and the `singularity.cacheDir` setting should be used to prevent this if that's the case.

Sorry @wilkb777 -   'quick'  was also rushed so I wasn't very clear (bad phrasing on my part)

e.g. 
https://github.com/nf-core/configs/blob/cheaha_singularity_update/docs/cheaha.md

I meant simple that it's clear for a user who doesn't look in the config itself but just the documentation, that many of the sometimes large temporary files will be indeed stored in the users 'space' rather than in `/tmp` (basically: no surprises for the user :) ). Even if `$USER_SCRATCH` is indeed the place to put these things in the case of your cluster, it's better for users to know when their 'environment' is being 'modified' (in a way).

There is nothing wrong with the idea nor implementation!

wilkb777:
😂  wow I totally went deep in the wrong direction there! Not a problem, thank you for the clarifications, they absolutely make sense! Will push changes to docs up with notes on the use of user scratch space.

jfy133 (approved):
Haha no worries, I was a tad confused at your initial response and was worried I had hit some hornets nest/back story with no context 😅

What you've written is perfect. You can merge!

wilkb777:
Oh, nothing with NF-Core or Nextflow so far to speak of! Snakemake on the other hand... 😬. It can be really tough writing with a tone of true curiosity and wanting to learn b/c I'm still getting up to speed on NF and NF-Core and just want to make sure I'm not missing anything. I also wanted to make sure I had completely thought threw everything and wasn't leaving the heavy lifting to someone else who may have more experience but way more responsibility, and therefore less time, with a massive project in NF & NF-Core. 

I'm super thankful for the work being done in this project, it's made life quite nice.

Thanks for the approval!

jfy133:
uh oh @wilkb777 @lianov I meant merge once tests pass (sorry for implicit assumption) - $USER_SCRATCH is failing in tests because of course it doesn't exist on the github runners (currently this is causing other tests to fail on other PRs :\)

jfy133:
hopefluly this should fix it: https://github.com/nf-core/configs/pull/410

lianov:
@jfy133 ahh that makes sense! We were wondering but thought that since we got the all clear for merging that it was good. To me your changes make sense for defining that variable earlier on. We will test it locally in our cluster an ensure it's not breaking anything. Thanks!!

lianov:
I will note that we may need to tweak a few things with `TMPDIR`, but a test will give us more insight. If any changes are needed we will follow-up with another PR @atrull314 @wilkb777

PR #407, opened 2022-08-25
Improve AWS-related config for Sage profile

BrunoGrandePhD:
Over the past few months, we've discovered several config optimizations for running our workflows on AWS Batch. Because the resource allocation tweaks haven't been tested thoroughly, I'm disabling them here with the intent of enabling them later once they're tested.

### DIFF
--- a/conf/sage.config
+++ b/conf/sage.config
@@ -1,62 +1,88 @@
+// Config profile metadata
 params {
   config_profile_description = 'The Sage Bionetworks profile'
   config_profile_contact = 'Bruno Grande (@BrunoGrandePhD)'
   config_profile_url = 'https://github.com/Sage-Bionetworks-Workflows'
 }
 
+// Leverage us-east-1 mirror of select human and mouse genomes
+params {
+  igenomes_base = 's3://sage-igenomes/igenomes'
+}
+
+// Enable retries globally for certain exit codes
 process {
-
-  cpus   = { check_max( 1    * slow(task.attempt), 'cpus'   ) }
-  memory = { check_max( 6.GB * task.attempt,       'memory' ) }
-  time   = { check_max( 24.h * task.attempt,       'time'   ) }
-
   errorStrategy = { task.exitStatus in [143,137,104,134,139,247] ? 'retry' : 'finish' }
   maxRetries    = 5
   maxErrors     = '-1'
-
-  // Process-specific resource requirements
-  withLabel:process_low {
-    cpus   = { check_max( 4     * slow(task.attempt),  'cpus'   ) }
-    memory = { check_max( 12.GB * task.attempt,        'memory' ) }
-    time   = { check_max( 24.h  * task.attempt,        'time'   ) }
-  }
-  withLabel:process_medium {
-    cpus   = { check_max( 12    * slow(task.attempt), 'cpus'   ) }
-    memory = { check_max( 36.GB * task.attempt,       'memory' ) }
-    time   = { check_max( 48.h  * task.attempt,       'time'   ) }
-  }
-  withLabel:process_high {
-    cpus   = { check_max( 24    * slow(task.attempt), 'cpus'   ) }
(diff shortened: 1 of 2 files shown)


BrunoGrandePhD:
Requesting a review from @nf-core/configs-team. Thanks!

jfy133 (changes requested):
1. If you don't need it, you should remove the resource allocation tweaks sectin - this should really be done on a per-pipeline basis
2. Related to 1 - you shouldn't need the `check_max` function as again should already be in each pipeline's `nextflow.config`

BrunoGrandePhD:
Thanks for the review, @jfy133! 

> 1. If you don't need it, you should remove the resource allocation tweaks sectin - this should really be done on a per-pipeline basis

I opted to uncomment that section and simplify it instead by removing the Sarek-specific section. I agree that optimizations should be done at the pipeline level (you can expect future PRs on this), but we also have some general tweaks that should apply to all pipelines. For example, we want to universally increase the time limits on jobs because it sucks when a long-running job gets killed prematurely. That's costly for a non-profit like us. 

> 2. Related to 1 - you shouldn't need the `check_max` function as again should already be in each pipeline's `nextflow.config`

As I mention in the comment above the `check_max` function, it isn't present in Sarek 2.7. Instead, they have a `check_resource` [function](https://github.com/nf-core/sarek/blob/e8f56e5bbb6f8a34793a6b5a2945981eb43090aa/nextflow.config#L255-L269). I also [tried](https://github.com/nf-core/configs/commit/6f0d9e6c43b8ee0373011f32d4dbf3c17811cd78) including this function in the Sarek-specific config, but it didn't work, presumably because of the evaluation order.

thomasyu888 (approved):

conf/sage.config
```diff
@@ -1,19 +1,52 @@
...
+params {
+  igenomes_base = 's3://sage-igenomes/igenomes'
+  max_memory    = '128.GB'
+  max_cpus      = 16
+  max_time      = '240.h'
+}
+
+// Enable retries globally for certain exit codes
+process {
+  errorStrategy = { task.exitStatus in [143,137,104,134,139,247] ? 'retry' : 'finish' }
```
BrunoGrandePhD, line 18: I wonder if there's a way to use this same strategy to adjust whether to increase the time allocation or memory based on the exit code. For example, on AWS Batch, exit code 143 means that the job reached its time limit.

jfy133:
> Thanks for the review, @jfy133!
> 
> > 1. If you don't need it, you should remove the resource allocation tweaks sectin - this should really be done on a per-pipeline basis
> 
> I opted to uncomment that section and simplify it instead by removing the Sarek-specific section. I agree that optimizations should be done at the pipeline level (you can expect future PRs on this), but we also have some general tweaks that should apply to all pipelines. For example, we want to universally increase the time limits on jobs because it sucks when a long-running job gets killed prematurely. That's costly for a non-profit like us.

OK fair enough - then lets leave it until you start getting the pipeline specific configs ready

> > 2. Related to 1 - you shouldn't need the `check_max` function as again should already be in each pipeline's `nextflow.config`
> 
> As I mention in the comment above the `check_max` function, it isn't present in Sarek 2.7. Instead, they have a `check_resource` [function](https://github.com/nf-core/sarek/blob/e8f56e5bbb6f8a34793a6b5a2945981eb43090aa/nextflow.config#L255-L269). I also [tried](https://github.com/nf-core/configs/commit/6f0d9e6c43b8ee0373011f32d4dbf3c17811cd78) including this function in the Sarek-specific config, but it didn't work, presumably because of the evaluation order.

Sarek :shakesfist: OK fair enough

jfy133 (approved):

conf/sage.config
jfy133: That's a very good question actuallly... might be worth asking on slack, it wouldn't surprise me if someone hasn't already had that issue.

jfy133:
You can ignore the 'repository root' test failure btw

PR #410, opened 2022-09-01
Fix cheaha config due to use of environmental variable

jfy133:
This fixes issues such as https://github.com/nf-core/configs/actions/runs/2949875596


jfy133:
@wilkb777 @lianov please can you test this, this should fix what I reported in your previous PR

maxulysse (approved):

jfy133:
@wilkb777 @lianov we will merge this now as to not block others, but please report back if it's not working correctly and we can find a better solution

PR #412, opened 2022-09-02
Fix cheaha config due to use of incorrect variable

atrull314:
---
Name: Fix cheaha config due to use of incorrect variable
---
**Description**: Updating the cheaha config to correct which variable is being assigned to TMPDIR. This should wrap up the changes and fixes starting in #404 and continued in #410.

### DIFF
--- a/conf/cheaha.config
+++ b/conf/cheaha.config
@@ -9,7 +9,7 @@ params {
 }
 
 env {
-    TMPDIR="$USER"
+    TMPDIR="$scratch_dir"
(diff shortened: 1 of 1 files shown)


atrull314:
Is there anything we can do to resolve the StackOverflowError that's failing?

jfy133:
> Is there anything we can do to resolve the StackOverflowError that's failing?

No, we are ignoring this for now -it'll be fixed in the next version of NXF apparently!

jfy133 (approved):

PR #428, opened 2022-10-13
vsc_ugent: limit queuesize

matthdsm:
Update vsc_ugent config to limit queueSize, so we stop getting slurm controller timeouts

### DIFF
--- a/conf/vsc_ugent.config
+++ b/conf/vsc_ugent.config
@@ -7,9 +7,11 @@ workDir = "$scratch_dir/work"
 // Perform work directory cleanup when the run has succesfully completed
 // cleanup = true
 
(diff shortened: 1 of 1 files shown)


nvnieuwk (approved):
Fingers crossed this will work :crossed_fingers:

matthdsm:
It does, I'm using it right now 😛

nvnieuwk:
It always works at first (it can be a bit sneaky)

matthdsm:
@nf-core/configs-team, how do we fix the failing CI? Or do we just ignore it?

nvnieuwk:
It's a known error with Nextflow, nothing we can do anything about for now
It should be fixed in the next Nextflow version

matthdsm:
allright, merging! If it breaks anything I blame @nvnieuwk!

PR #457, opened 2023-02-10
Added Azure Batch pools configuration file for rnaseq pipeline

olesya13:
**Name:** Azure Batch config for nf-core/rnaseq pipeline
**About:** defines several Azure Batch pools with virtual machines from Edv4-series and assigns pools to labels according to the requirements defined in pipeline's base.config

### DIFF
--- a/conf/pipeline/rnaseq/azurebatch_pools_Edv4.config
+++ b/conf/pipeline/rnaseq/azurebatch_pools_Edv4.config
@@ -1,0 +1,61 @@
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    nf-core/rnaseq Nextflow config file for Azure Batch pools
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Defines several Azure Batch pools with virtual machines from Edv4-series, 
+    assigns pools to labels according to the requirements defined in base.config.
+    You might need to adjust vmCount and maxVmCount depending on your Batch account quotas.
+
+    Use as follows:
+        nextflow run nf-core/rnaseq -c '<Path to this configuration file>'  -profile azurebatch
+            --input 'az://<Path to input file>' --outdir 'az://<Output directory>' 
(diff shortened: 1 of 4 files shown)


abhi18av (approved):
Technilcally, looks good to merge ✅ 

But I'm not sure if this would solve the issue of automated testing on the nf-core side or would facilitate beginner users as new accounts don't have access to `EDv4` VMs 

image

conf/pipeline/rnaseq/azurebatch_pools_Edv4.config
```diff
@@ -0,0 +1,61 @@
...
+
+params {
+    config_profile_description = 'Azure Batch Edv4 pools profile'
+    config_profile_contact = 'Olesya Melnichenko (@olesya13) & Venkat Malladi (@vsmalladi)'
+    config_profile_url = 'https://azure.microsoft.com/services/batch/'
+}
+
+process {
+    queue = 'Standard_E2d_v4'
+    withLabel:process_low   {queue = 'Standard_E2d_v4'}
```
Emiller88, line 24:
```suggestion
    withLabel:process_low {queue = 'Standard_E2d_v4'}
```
Might pick a lane on the alignment or no alignment here.

Emiller88 (approved):
Looks good to me!

maxulysse:
Isn't this something more for azure at a global level than just for the rnaseq pipeline?

vsmalladi:
> Isn't this something more for azure at a global level than just for the rnaseq pipeline?

Yes that is correct. The quota is at the global level.

olesya13:
@abhi18av Do you think it will be useful to add configuration file for Ddv4 series? We ran experiments on both Ddv4 and Edv4, so I have Ddv4 config ready. Edv4 turned out to be more cost efficient in our experiments, that's the main reason why I decided to add this one.

maxulysse:
Can you add it in the main README too please:
https://github.com/nf-core/configs/blob/master/README.md#pipeline-specific-institutional-documentation

conf/pipeline/rnaseq/azurebatch_pools_Edv4.config
olesya13: fixed - removed extra spaces

olesya13:
@maxulysse documentation added

abhi18av:
Whoops, dropped the ball on this one - @olesya13 , I'd say that this config is enough for the initial work since on the nf-core Azure account we are capped on 200 low-priority CPUs therefore it should work with that tweak.

Though, later on we'd ideally also test with the `Ddv4` VMs since that particular series has been recommended in the past - https://github.com/seqeralabs/nf-core-tower-azure-poc

PR #462, opened 2023-02-16
Update scw.md

marccamb:
---
name: SCW documentation update  
about: Update on the config documentation for Super Computing Wales
---

Included instructions for a local installation of mamba to fix problems with the current conda installation on SCW and allow Singularity container download


prototaxites:
@nf-core-bot fix linting

docs/scw.md
```diff
@@ -11,27 +11,31 @@ module avail
 module load nextflow/21.10.6
 ```
```
prototaxites, line 12: I think if we're asking users to do a conda install for the nf-core tools, we might just direct them to use the conda version of Nextflow as well - that way they can stay up-to-date more easily, too.

maxulysse (approved):

PR #484, opened 2023-04-06
adding vsc_kul_uhasselt configs

Joon-Klaps:
---
name: vsc_kul_uhasselt
about: A new cluster config for the KULeuven and Uhasselt Tier2 clusters Genius and Wice
---
The VSC (Flemish supercomputer) has multiple clusters which are specific to flemish universities.
I propose not to extend the `vsc_ugent.config` file but instead make a new config `vsc_kul_uhasselt.config` as some of the prefixed variables like `$VSC_SCRATCH_VO_USER` are not available on KUL's and UHasselt's HPC. 

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,6 +109,7 @@ jobs:
           - "utd_sysbio"
           - "uzh"
           - "vai"
+          - "vsc_kul_uhasselt"
           - "vsc_ugent"
           - "wehi"
           - "xanadu"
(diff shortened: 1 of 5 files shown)


Joon-Klaps:
The Genius and Wice clusters are currently under maintenance so they haven't been tested yet

nvnieuwk:
There are only two clusters on both HPCs? The config looks good to me!

Joon-Klaps:
> There are only two clusters on both HPCs? The config looks good to me!

You're right there is one more the [`Superdome`](https://docs.vscentrum.be/en/latest/leuven/tier2_hardware/superdome_hardware.html) which is part of the Genius cluster.

Joon-Klaps:
I figured out what was wrong, the important difference between vsc_ugent and vsc_kul_uhasselt is that it requires a credential account. 
I tested all 3 configurations and tested with resume, all were successful.

nvnieuwk (approved):
LGTM! Did you manage to run a full run with this config?

PR #507, opened 2023-06-05
Re-add cpu/memory specifications to MAG @ EVA

jfy133:
As we had new withName process definitions, these were entirely overwriting any other settings (whereas we only wanted to update a single line) - so while the time was updated, all other resource requirements resorted to a default of 1 CPU.

In this case I've copied and pasted the entire MAG base.config, but just replacingf the `time` definitions.

### DIFF
--- a/conf/pipeline/mag/eva.config
+++ b/conf/pipeline/mag/eva.config
@@ -20,21 +20,30 @@ process {
     }
 
     withLabel:process_single {
+        cpus   = { check_max( 1                  , 'cpus'    ) }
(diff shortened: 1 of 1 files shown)


jfy133:
https://tower.nf/orgs/mpi-archaeogenetics/workspaces/biocluster-monitoring/watch/4EkaoVOqZLq6WU

alexhbnr (approved):
The changes seem to work fine now and the SGE picks up the correct number of CPUs and amount of memory as specified here.

PR #509, opened 2023-06-10
[WIP] Yale McCleary config

rhagenson:
---
name: McCleary
about: A config for the Yale McCleary cluster
---


rhagenson:
This profile is a WIP as I am sure there are at least a handful of oversights or better ways to configure usage. The corresponding documentation for the cluster itself is found here: https://docs.ycrc.yale.edu/clusters/mccleary/

I have attempted to have a quasi dual-config where jobs that are labeled get reasonable defaults, while jobs that are not labeled are automatically queued to the correct location.

Not a blocker for submission, but I would like to add automatic usage of:

- Add `bigmem` to config (automatic, much like the usage of `day` or `week` partitions)
- Add `ycga` and `ycga_bigmem` to config (opt-in, for when a workflow is using data from YCGA)

rhagenson:
Something is definitely buggy about this config as-is. I attempted to run nf-core/sarek with it (and the test data) and it fails with a stack overflow for `NFCORE_SAREK:SAREK:PREPARE_INTERVALS:CREATE_INTERVALS_BED` -- removing the `mccleary` profile from the submission and it no longer fails at that early stage.

docs/mccleary.md
```diff
@@ -0,0 +1,35 @@
...
+To use, run the pipeline with `-profile mccleary`. This will download and launch the [`mccleary.config`](../conf/mccleary.config) which has been pre-configured with a setup suitable for the McCleary cluster. Using this profile, a docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.
+
+## Below are non-mandatory information e.g. on modules to load etc
+
+Before running the pipeline you will need to load Java 11 using the environment module system on McCleary. You can do this by issuing the commands below:
+
+```bash
+## Load Java 11 environment modules
+module purge
+module load Java/11
```
ggabernet, line 14:
```suggestion
module load Java/17.0.4
```

ggabernet: Java 17 also worked for me, maybe better to add the latest version then

conf/mccleary.config
```diff
@@ -0,0 +1,118 @@
...
+    executor {
+      submitRateLimit = '200/1h'
+    }
+
+    singularity {
+      enabled = true
+    }
+
+    params {
+      igenomes_base = '/gpfs/gibbs/data/genomes/'
```
ggabernet, line 17: Awesome you added this!

docs/mccleary.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Recommendation on environment variables to set
+
+Although not required, it is recommended to set certain Nextflow specific environment variables to make better use of cluster resources.
+
+```bash
+NXF_OPTS="-Xms500M -Xmx8G"
+NXF_WORK="$PALMER_SCRATCH"
+NXF_SINGULARITY_CACHEDIR="$PALMER_SCRATCH"
+```
```
ggabernet, line 25: Not sure whether I would recommend users to set up environment variables, sometimes one forgets they are there and it's hard to debug. Especially the 1st one I would not recommend, and I like directly specifying workdirs for each project

rhagenson: Perhaps we can change the language around the recommendation? The core reason for recommending `WORK` and `SINGULARITY_CACHEDIR` is to put all the intermediate files (tons of data) into a location that will be automatically cleaned up by YCRC.

The reason for recommending `OPTS` is from https://www.nextflow.io/blog/2021/5_tips_for_hpc_users.html

---

I see no strong reason to keep the `OPTS` recommendation in place as the JVM should use available memory rather than need to be explicitly told how much heap to allocate. The recommendation to use `NXF_WORK=$PALMER_SCRATCH` I think is a strong recommendation if we frame the reason for it rather than just say "we recommend setting these variables" -- what are your thoughts?

rhagenson:
@ggabernet What are your thoughts on the overall design of the config? Does the dual-purpose make sense to you? Have some settings that apply themselves when a process is unlabeled and others that apply when labeled?

docs/mccleary.md
ggabernet: Sounds good in principle, and it is a preference thing really, I like providing the work directory for each run directly with  the nextflow parameter `-w palmer_scratch/project_x/work` for each project separately, that way I do not have a mix up of work directories and I can go and debug a run if I need to inside the workdir. For the other two I guess the env variable is fine. In any case setting the `-w` parameter overrides the env variable.

conf/mccleary.config
ggabernet: oh yes I was involved in the discussions, that's not happening any time soon :)
but yes, we should check that the data is regularly updated, I will try to figure out how this is done

ggabernet:
The logic of the config looks good to me, just had one comment but not so critical. Did you test it for some runs and is it working? Then we can merge and I can test it again. I did try setting also some SLURM time limits in some of our configs, did you ever hit any job submission rate limit?

conf/mccleary.config
```diff
@@ -0,0 +1,118 @@
+params {
+  config_profile_description = 'Yale YCRC Configurations'
+  config_profile_contact = 'Ryan A. Hagenson'
+}
+
+profiles {
+  mccleary {
```
ggabernet, line 7:
```suggestion
```

ggabernet: Ah I see now, the profile is not defined here, it is defined within the `nfcore_custom.config` file, so you can removed those lines and the indentation

conf/mccleary.config
```diff
@@ -0,0 +1,118 @@
...
+      // Do not bother checking max_time, set above 24.h
+      withLabel:process_high {
+        cpus = { check_max( 12 * task.attempt, 'cpus' ) }
+        memory = { check_max( 120.GB * task.attempt, 'memory' ) }
+        time = { 24.hours * task.attempt < 7.days ? 24.hours * task.attempt : 7.days}
+        queue = 'week'
+      }
+    }
+  }
+}
```
ggabernet, line 85:
```suggestion
```

rhagenson:
@ggabernet I finally had time to test this config again as it was causing some job failures that I was not able to track down. The underlying issue was the combination of

```yaml
queue = {
  switch (task.time) {
    case { it > 1.day }:
      return 'week'
    default:
      return 'day'
  }
}
```

and

```yaml
time = { task.time < 7.days ? task.time : 7.days}
```

I removed the latter `time` assignment and updated the `params.max_time` accordingly. I have tested this new config and it works splendidly.

I also simplified the config where I could by using ternary operators on single lines.

ggabernet:
Amazing, let's get this out. I will add some other suggestions for issues I've experienced

conf/mccleary.config
ggabernet: hmm, I'm unsure that this will work, as it does not seem to be an exact copy of the nf-core AWS igenomes base, so it's missing some of the files that we have there. Do you happen to know who handles the reference data there?

conf/mccleary.config
```diff
@@ -0,0 +1,91 @@
+params {
+  config_profile_description = 'McCleary Cluster at Yale'
+  config_profile_contact = 'Ryan A. Hagenson'
+}
+
+// Documented job rate limit
+executor {
+  submitRateLimit = '200/1h'
```
ggabernet, line 8:
```suggestion
  submitRateLimit = '199/1h'
```

docs/mccleary.md
```diff
@@ -0,0 +1,35 @@
...
```
ggabernet, line 24:
```suggestion


There is currently a Batch file size submission limit in the McCleary cluster, if you're running into the error `sbatch: error: Batch job submission failed: Pathname of a file, directory or other parameter too long`, you can set the environment variables:

```bash
export NXF_WRAPPER_STAGE_FILE_THRESHOLD='40000'
```
This will only work for Nextflow versions >= 23.10.0 so you might need to update Nextflow first.

```

File: conf/mccleary.config (hunk 2 above)
Comment (rhagenson) [reply conf/mccleary.config]: I do not. The [documentation](https://docs.ycrc.yale.edu/clusters/mccleary/#public-datasets) just states a general contact for any updates/requests.

**Review by ggabernet — COMMENTED** [review]
I just have some comments, happy to discuss

File: conf/mccleary.config [hunk 8]
```diff
@@ -0,0 +1,91 @@
...
+
+process {
+  executor = 'slurm'
+  cache = 'lenient'
+  scratch = true
+
+  // Update amount of max retries and set "retry" as the error strategy for all error codes
+  errorStrategy = 'retry'
+  maxRetries = 5
+  maxErrors = '-1'
```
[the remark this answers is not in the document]
ggabernet: Or are they needed here due to the `scavenge` queue, in case they get killed?

rhagenson:
@ggabernet Why did you merge `nf-core/configs:master` into my branch? The changes on `master` are unrelated to this PR.

conf/mccleary.config
ggabernet: ok, then we should remove this parameter until we have somewhere an exact copy of the nf-core igenomes reference data, as this one won't work well

ggabernet:
@rhagenson because there were some merge conflicts, if you check the diff you will see that only your changes are showing

rhagenson:
You found merge conflicts when merging commits that are unrelated to this PR? This PR now includes a HEAD commit that changes 113 files with thousands of lines changes. While yes it is true that the GitHub UI diff shows my changes, the vast majority of the changes included in this PR are not mine nor are they related to the McCleary configuration. I am not going to proceed with a PR that has so many related changes in it. Please revert your merge commit so that I may rebase onto `nf-core/configs:master` so that the changes in this PR are related to adding the McCleary configuration.

ggabernet:
Hmm I don't see the changes you mention, the GitHub UI only shows your diff, and resolving conflicts like this and updating with the master branch is something we do all the time.

jfy133:
Hi @rhagenson and @ggabernet .

I see there is already an existing config merged in/updated last week via a separate PR

Can this one now be closed? If I don't hear back I will close next week.

jfy133:
Closing now, feel free to re-open if relevant.

PR #510, opened 2023-06-19
Updating VSC_KUL_UHASSELT config for overhead

Joon-Klaps:
Current config, was setting the maximum allowed RAM to the total available RAM of a node leaving nothing for overhead. This resulted in an instantly failing of the jobs. This was fixed by leaving a large amount (22GB - 36GB - 48GB) for overhead depending on the total available RAM of the node. Some logic was also added to submit to larger memory nodes on both clusters. 

- [X] Test the config

### DIFF
--- a/conf/vsc_kul_uhasselt.config
+++ b/conf/vsc_kul_uhasselt.config
@@ -51,8 +51,8 @@ profiles {
         }
 
         process {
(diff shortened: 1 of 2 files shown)


maxulysse (approved):

nvnieuwk (approved):
LGTM

PR #518, opened 2023-06-30
CRUKMI config update

SPPearce:
Set the queue automatically and lower the maximum amount of RAM that can be requested on the standard compute cores.


ewels (approved):
LGTM 👍🏻

PR #519, opened 2023-07-01
Ebi

saulpierotti-ebi:
---
name: EBI config
about: A new cluster config
---

I followed the required steps and added 2 config files, one for the LSF login nodes of the EBI codon cluster, and one for the SLURM login nodes.


conf/ebi_codon.config
```diff
@@ -0,0 +1,60 @@
...
+    enabled = true
+    // the default is 20 minutes and fails with large images
+    pullTimeout = "3 hours"
+}
+
+conda {
+    enabled = true
+    // use the faster mamba solver instead of conda
+    useMamba = true
+}
```
maxulysse, line 26: enabling both?

saulpierotti-ebi: I use it like this usually so that I can run some processes with conda and some with singularity when developping my own pipelines. When running nf-core pipeline I usually use singularity enabling the appropriate profile. Do you think that it would it be better to keep just singularity in the config?

docs/ebi_codon.md
```diff
@@ -1,3 +1,5 @@
+<<<<<<< HEAD
+=======
```
maxulysse: this looks bad

maxulysse:
what happend to your md files?

maxulysse:
sorry for the mess with prettier

saulpierotti-ebi:
Seems fine now

maxulysse:
I'm just concerned by the conda and singularity enabled both at the same time.
@jfy133 what's your take on that?

jfy133:
Huh weird,I thought I had left a comment already about that. I agree don't have both

This should be a production profile so you should just stick with singularity here, you can always have a sub profile for development if you really need (or just a local config that overrides the software environment scope)

saulpierotti-ebi:
Yes makes sense, I will remove conda then

conf/ebi_codon.config
```diff
@@ -0,0 +1,54 @@
...
+    config_profile_contact = "Saul Pierotti (@saulpierotti-ebi)"
+    config_profile_description = "The European Bioinformatics Institute HPC cluster (codon) profile"
+    config_profile_url = "https://www.ebi.ac.uk/"
+}
+
+singularity {
+    enabled = true
+    // the default is 20 minutes and fails with large images
+    pullTimeout = "3 hours"
+}
```
maxulysse, line 20: I'll advise in adding a cacheDir for singularity

maxulysse: I see that you recommend in the docs that the location should be
```
"/hps/software/users/<group>/<user_id>/nextflow_software_cache/singularity"
```

I'd reckon it'll be more efficient if shared with all

saulpierotti-ebi: There is not a directory writable by everyone in our cluster...

maxulysse: ok, then nevermind

maxulysse (approved):

PR #522, opened 2023-07-05
Add tier 1 clusters for VSC UGent

nvnieuwk:


maxulysse:
Don't you need to add this tier1_project params to the schema ignore?

nvnieuwk:
Good point! I'll add it, but it should be a required parameter when running with the tier1 clusters so I also have to find a way to enforce that

maxulysse:
> Good point! I'll add it, but it should be a required parameter when running with the tier1 clusters so I also have to find a way to enforce that

cf https://github.com/nf-core/configs/blob/8105c09b61c9312a7a78015098a953b862de14d7/conf/uppmax.config#L9

nvnieuwk:
@maxulysse tests are still failing because the error message produced by the config, what should I do about this?

conf/vsc_ugent.config
```diff
@@ -4,6 +4,19 @@ if (!params.tier1_project) {
     params.tier1_project = null
 }
 
+// Get the hostname and check some values for tier1
+def hostname = "doduo"
```
maxulysse, line 8: pokemon fan?

maxulysse (approved):

conf/vsc_ugent.config
nvnieuwk: Hehehehe maybe ;) all our clusters are named after pokemon

maxulysse: yeah, I noticed dodrio too

PR #529, opened 2023-07-18
Create CODEOWNERS file

Pandapip1:
This will automatically request @nf-core/maintainers for review for any PR opened.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @nf-core/maintainers


jfy133:
I'm not sure we want this by _default_, right @maxulysse ? we only require that in the checklist for people new to the repo. In some cases people have their own colleagues/collaborators outside of the maintainers team who will do the review on their behalf.

maxulysse:
> I'm not sure we want this by _default_, right @maxulysse ? we only require that in the checklist for people new to the repo. In some cases people have their own colleagues/collaborators outside of the maintainers team who will do the review on their behalf.

Yeah, I kinda like the idea, but maintainers work more with modules and pipelines anyway. I feel it'll make more noise than he actually useful

PR #537, opened 2023-08-15
Remove crukmi profile from sarek specific pipelines

SPPearce:
When I edited the crukmi profile for our new cluster, I removed the sarek specific pipeline config, but forgot to remove it from the list of sarek configs.


maxulysse (approved):

SPPearce:
Thanks Maxime!

PR #547, opened 2023-09-12
Ethz euler

jpadesousa:
---
name: New Config
about: A new cluster config
---

- [x] Requested review from @nf-core/maintainers and/or #request-review on slack

### DIFF
--- a/conf/ethz_euler.config
+++ b/conf/ethz_euler.config
@@ -27,7 +27,6 @@ process {
         scratch             = 'true'
         errorStrategy       = 'retry'
         maxRetries          = 3
-        maxErrors           = '-1'
         beforeScript        = 'module load eth_proxy'
 }
 
--- a/conf/ethz_euler.config
+++ b/conf/ethz_euler.config
@@ -1,0 +1,39 @@
+/*
+* ----------------------------------------------------------------
+*   nf-core pipelines config file for the ETH Zurich Euler cluster
+* ----------------------------------------------------------------
+*/
+
+params {
+        config_profile_name        = 'ethz_euler'
+        config_profile_description = 'ETH Zurich Euler cluster profile -- provided by nf-core/configs'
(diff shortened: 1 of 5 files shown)


conf/ethz_euler.config
```diff
@@ -0,0 +1,51 @@
...
+executor {
+        name                = 'slurm'
+        perCpuMemAllocation = true
+}
+
+process {
+        scratch             = 'true'
+        errorStrategy       = 'retry'
+        maxRetries          = 3
+        maxErrors           = '-1'
```
pontus, line 30: What is the intention of this? https://www.nextflow.io/docs/latest/process.html#maxerrors doesn't document strings or -1 (but says disabled by default).

jpadesousa: Hi @pontus,
Yes, you are right. That can be removed.

conf/ethz_euler.config
```diff
@@ -0,0 +1,51 @@
...
+        beforeScript        = 'module load eth_proxy'
+}
+
+singularity {
+        enabled    = true
+        autoMounts = true
+        cacheDir   = "${APPTAINER_CACHEDIR}"
```
pontus, line 36:
```suggestion
```

This doesn't really make sense to me - what apptainer uses `APPTAINER_CACHEDIR` for is quite different from what nextflow uses it for (e.g. for apptainer you really don't want it shared whereas for the nextflow case it makes sense to keep them shared within the process directory).

I'm also not certain it's easier to set `APPTAINER_CACHEDIR` than `NXF_SINGULARITY_CACHEDIR` and the default behaviour of keeping them in work if not set seems more sensible than what I guess will happen here if unset.

jpadesousa: Thank you for the input. You are right. We have already NXF_SINGULARITY_CACHEDIR  defined once the user loads the nextflow module. So this would just require and extra unnecessary step. I will remove it.

pontus (approved):

jfy133 (approved):

PR #551, opened 2023-09-15
FEAT: Add software_license profile

maxulysse:
to simplify GHA with Sentieon for pipelines and modules


.github/workflows/main.yml
```diff
@@ -92,6 +92,7 @@ jobs:
           - "mpcdf"
           - "munin"
           - "nci_gadi"
+          - "nfcore_license"
```
adamrtalbot, line 95: I don't like the name `nfcore_license`. What does that mean? A license to use nf-core? 

Call it `sentieon_license_server_extension` if you have to use have a separate one. Ideally, we just have one `-profile sentieon` which handles both of these situations.

adamrtalbot (changes requested):

.github/workflows/main.yml
maxulysse: yeah wanted to have a simple and short name

conf/nfcore_license.config
```diff
@@ -0,0 +1,13 @@
+//Profile config names for nf-core/configs
+params {
+    config_profile_description = 'Sentieon license set up for nf-core by nf-core/configs'
+    config_profile_contact = 'Maxime U Garcia (@maxulysse)'
+    config_profile_url = 'https://www.sentieon.com'
+}
+
+process {
+    withLabel: 'sentieon' {
+        ext.sentieon_auth_mech_base64 = secrets.SENTIEON_AUTH_MECH_BASE64 ?: ''
```
asp8200, line 10: `secrets.SENTIEON_AUTH_MECH_BASE64 ?: ''` failed for Ram when `secrets.SENTIEON_AUTH_MECH_BASE64` wasn't set.

asp8200: So the idea is to have many different licenses in `conf/nfcore_license.config`? Say `nfcore_license.config` refers to nf secrets for tool X and tool Y but the user only has license for X and can only set the nf secrets for X, then it seems to me that this will fail for the same reasons that `test_cache` is failing when the user doesn't have a test-license for Sentieon.

asp8200, line 10:
```suggestion
        ext.sentieon_auth_mech_base64 = secrets.SENTIEON_AUTH_MECH_BASE64
```

conf/nfcore_license.config
```diff
@@ -0,0 +1,13 @@
...
+        ext.sentieon_auth_data_base64 = secrets.SENTIEON_AUTH_DATA_BASE64 ?: ''
```
asp8200, line 11:
```suggestion
        ext.sentieon_auth_data_base64 = secrets.SENTIEON_AUTH_DATA_BASE64
```

conf/nfcore_license.config
maxulysse: yeah, this should only works when we set up secrets when we start the tests, similarly to how we do it within our CI

.github/workflows/main.yml
adamrtalbot: Not nfcore_license, pick anything else!

conf/nfcore_license.config
adamrtalbot: Do we know this does or doesn't work? I know Ram tested it but I think we were flailing around trying different solutions.

adamrtalbot: It's nothing to do with a 'test license' so we can't call it 'test'.

conf/software_license.config
```diff
@@ -0,0 +1,13 @@
...
+    config_profile_contact = 'Maxime U Garcia (@maxulysse)'
+    config_profile_url = 'https://nf-co.re'
+}
+
+process {
+    withLabel: 'sentieon' {
+        ext.sentieon_auth_mech_base64 = secrets.SENTIEON_AUTH_MECH_BASE64 ?: ''
+        ext.sentieon_auth_data_base64 = secrets.SENTIEON_AUTH_DATA_BASE64 ?: ''
+    }
+}
```
adamrtalbot, line 13: Wrap this behind a parameter:
```
params.sentieon_extension = false
```

```suggestion
if ( params.sentieon_extension ) {
    process {
        withLabel: 'sentieon' {
            ext.sentieon_auth_mech_base64 = secrets.SENTIEON_AUTH_MECH_BASE64 ?: ''
            ext.sentieon_auth_data_base64 = secrets.SENTIEON_AUTH_DATA_BASE64 ?: ''
        }
    }
}
```

docs/software_license.md
```diff
@@ -0,0 +1,5 @@
+# nf-core/configs: Software licenses set up for nf-core
+
+To use, run the pipeline with `-profile nfcore_license`. This will download and launch
+the [`nfcore_license.config`](../conf/nfcore_license.config) which has been pre-configured with a
+setup suitable for using _Sentieon_ within the nf-core GHA.
```
adamrtalbot, line 5: Correct profile name

adamrtalbot:
```suggestion
To use, run the pipeline with `-profile software_license`. This will download and launch
the [`software_license.config`](../conf/software_license.config) which has been pre-configured with a
setup suitable for using _Sentieon_ within the nf-core GHA.
```

conf/nfcore_license.config
maxulysse: let's merge without and we can improve on later, that's the whole point behind this implementation

nfcore_custom.config
```diff
@@ -85,10 +85,11 @@ profiles {
     rosalind_uge     { includeConfig "${params.custom_config_base}/conf/rosalind_uge.config" }
     sage             { includeConfig "${params.custom_config_base}/conf/sage.config" }
     sahmri           { includeConfig "${params.custom_config_base}/conf/sahmri.config" }
-    sanger           { includeConfig "${params.custom_config_base}/conf/sanger.config"}    
+    sanger           { includeConfig "${params.custom_config_base}/conf/sanger.config"}
     scw              { includeConfig "${params.custom_config_base}/conf/scw.config"}
     seawulf          { includeConfig "${params.custom_config_base}/conf/seawulf.config"}
     seg_globe        { includeConfig "${params.custom_config_base}/conf/seg_globe.config"}
+    software_license { includeConfig "${params.custom_config_base}/conf/software_license.config"}
```
sateeshperi, line 92: wouldn't it be better to be more explicit by calling it `sentieon_software_license.config` ?

maxulysse: the idea would be to be future proof and if later we need a licence for let's say parabricks, dragen, or anything else, we have a config ready for it to use at an nf-core level

sateeshperi (approved):
lgtm

PR #568, opened 2023-10-12
Add University of Dundee cluster config

dsloanm-uodit:
---
name: New Config
about: A new cluster config
---

A workaround for https://github.com/nextflow-io/nextflow/issues/2449 is included in the config file. Following the approach at https://github.com/nf-core/configs/blob/9c4830c/conf/hebbe.config#L16-L17

- [x] Requested review from @nf-core/maintainers and/or #request-review on slack

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -117,6 +117,7 @@ jobs:
           - "uge"
           - "unibe_ibu"
           - "unc_lccc"
+          - "uod_hpc"
           - "uppmax"
           - "utd_ganymede"
           - "utd_sysbio"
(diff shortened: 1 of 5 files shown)


jfy133 (approved):
Theoretically OK, however you've not specified any software environment system which is also quite important (conda, docker, singularity etc.). I see you ask people to use conda to install Nextflow itself, is that the only engine available? If so I would turn that on in the config too

dsloanm-uodit:
Thanks for the comments - I have explicitly enabled conda in the config. Ideally we would like to use Singularity but our compute nodes aren't quite set up for it yet. It's something we're aiming for, however.

jfy133:
One last thing to document: I would recommend users specifying a [NXF_CONDA_CACHEDIR](https://www.nextflow.io/docs/latest/config.html#environment-variables) env variable so they don't create the same conda environments for every run

dsloanm-uodit:
I've added some notes on NXF_CONDA_CACHEDIR (and fixed a typo in a section header). Thanks again

jfy133:
Will wait for tests to pass the nwill merge :+1:

PR #555, opened 2023-09-20
[WIP] Update sanger.config

DLBPointon:
Updating the sanger profile to take into account the memory limits when assigning queues. The current config is killing jobs as memory is not taken into account.

I have decided on the optimal resources per queue as although you can submit a 400GB job on normal, the priority becomes so low it will not run for a long time.

---
name: New Config
about: A new cluster config
---

Please follow these steps before submitting your PR:

- [x] If your PR is a work in progress, include `[WIP]` in its title
- [X] Your PR targets the `master` branch
- [ ] You've included links to relevant issues, if any
- [ ] Requested review from @nf-core/maintainers and/or #request-review on slack

If you require/still waiting for a review, please feel free to request from @nf-core/maintainers

Please see uploading to`nf-core/configs` for more details:
https://github.com/nf-core/configs#uploading-to-nf-coreconfigs

Thank you for contributing to nf-core!

PR #587, opened 2023-11-08
Updated the Sanger config to list more queues and cover multiple internal clusters

muffato:
Supersedes #555 

This config will first query the name of the cluster, to allow defining different values in the rest of the file depending on the cluster.

1. The list of queues has been expanded to include `hugemem` and `teramem`.
2. I've reduced the threshold of the `small` queue to 1 minute, as per our HPC guidelines.
3. The maximum resources can now be defined per cluster, and I've defined different values for `tol` and `farm5`.
4. For `farm5`, I've added `/data` to the list of mount points.
5. For the `tol` cluster, I've removed all the automatic mounting of `/lustre`, etc, because we want to make sure that all the inputs are properly defined in the pipelines. Pipelines that do require whole filesystems to be mounted need to define their own `singularity.runOptions`.

### DIFF
--- a/conf/sanger.config
+++ b/conf/sanger.config
@@ -36,8 +36,10 @@ process {
             "week"
         } else if ( task.time > 12.hour ) {
             "long"
-        } else {
+        } else if ( task.time > 1.min ) {
             "normal"
+        } else {
+            "small"
         }
     }
 
--- a/conf/sanger.config
+++ b/conf/sanger.config
@@ -1,14 +1,46 @@
+
+// Extract the name of the cluster to tune the parameters below
+def clustername = "farm5"
+try {
+    clustername = ['/bin/bash', '-c', 'lsid | awk \'$0 ~ /^My cluster name is/ {print $5}\''].execute().text.trim()
+} catch (java.io.IOException e) {
+    System.err.println("WARNING: Could not run lsid to determine current cluster, defaulting to farm")
+}
+
 // Profile details
 params {
-    config_profile_description = 'The Wellcome Sanger Institute HPC cluster (farm5) profile'
+    config_profile_description = "The Wellcome Sanger Institute HPC cluster (${clustername}) profile"
     config_profile_contact = 'Priyanka Surana (@priyanka-surana)'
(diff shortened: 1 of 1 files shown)


muffato:
ping @BioinfoTongLI who I can't add as "reviewer" for an unknown reason

DLBPointon:
This looks good to me @muffato, thanks for taking a look.

maxulysse (approved):
LGTM

muffato:
@maxozo @BioinfoTongLI could you take a look ? This will affect you

maxozo:
This looks good to me! Don’t think it will break anything for me. Thanks! On 10 Nov 2023, at 11:19, Matthieu Muffato ***@***.***> wrote:﻿
@maxozo @BioinfoTongLI could you take a look ? This will affect you

BioinfoTongLI:
As I am not an extensitve user of all the different queues. I am unsure if the if-else will capture all the use cases. Or is it trying to capture that at all?
I see that it is now not wrapped into `withlabel` and is default setting. Wouldn't it be safer to wrap this part into `withlabel cpu-only`?

muffato:
> As I am not an extensitve user of all the different queues. I am unsure if the if-else will capture all the use cases. Or is it trying to capture that at all?

The chain of if-else always has an _else_ and defines a queue for all scenarios.

> I see that it is now not wrapped into `withlabel` and is default setting. Wouldn't it be safer to wrap this part into `withlabel cpu-only`?

That's the way the configs work, I didn't change that. "CPU-only" is the default mode for processes and is captured by having the `queue = {...}` directly in the `process` block. Processes that need GPUs must add the `gpu` label and are then captured by the `withLabel: gpu { ... }` block which _overrides_ the default values set above. Moving the default queue settings to a `withLabel: cpu_only { ... }` block would mean that all existing pipelines would instantly break as they couldn't get their queue settings. I don't want to do that 😅

BioinfoTongLI:
thanks for the explanation @muffato! I think the `override` piece of inofrmation is the missing piece of my knowledge -  That sounds then! All good to me!

PR #592, opened 2023-11-17
new configuration for using nf-core pipelines at University of Luebeck

finjasygo:
---
name: New Config
about: A new cluster config
---

### DIFF
--- a/conf/uzl_omics.config
+++ b/conf/uzl_omics.config
@@ -16,10 +16,6 @@ process {
     scratch = '$SCRATCH'
 }
 
-workDir = "$WORK/nf-core_rnaseq_workdir"
-
 singularity {
     enabled = true
-    runOptions = "-B $WORK/nf-core_rnaseq_workdir -B $HOME/.nextflow/assets/nf-core/rnaseq/bin"
-    cacheDir = "$WORK/.singularity"
 }
--- a/conf/uzl_omics.config
+++ b/conf/uzl_omics.config
@@ -1,0 +1,25 @@
+params {
+    config_profile_description = 'University of Luebeck, OMICS cluster profile provided by nf-core/configs.'
+    config_profile_contact = 'Finja Sygo (@finjasygo) and Varun Sreenivasan (@varunkas) (Institut fuer Humangenetik)'
+    config_profile_url = 'https://www.itsc.uni-luebeck.de/dienstleistungen/omics.html'
+}
+
+params {
+    max_memory = 760.GB    
+    max_cpus   = 48
+    max_time   = 72.h     
+}
+
+process {
+    executor = 'slurm'
+    queue = 'shortterm'
+    scratch = '$SCRATCH'
+}
+
+workDir = "$WORK/nf-core_rnaseq_workdir"
+
+singularity {
+    enabled = true
+    runOptions = "-B $WORK/nf-core_rnaseq_workdir -B $HOME/.nextflow/assets/nf-core/rnaseq/bin"
+    cacheDir = "$WORK/.singularity"
+}
--- a/docs/uzl_omics.md
+++ b/docs/uzl_omics.md
@@ -1,0 +1,27 @@
+# nf-core/configs: UzL OMICS Cluster Configuration
+
+The rnaseq nf-core pipeline has been successfully configured for use on the UzL OMICS cluster at the University of Luebeck.
+Implementation of nf-core pipelines for use on the UzL OMICS is in process. 
+
+To use, run the pipeline with `-profile uzl_omics`. This will download and launch the `uzl_omics.config` which has been pre-configured with a setup suitable for the UzL OMICS cluster. Using this profile, a docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.
+
+Before running the pipeline you will need to load Nextflow and Singularity using the environment module system on UzL OMICS cluster. You can do this by issuing the commands below:
+
+```bash
+module load nextflow
+module load singularity
+```
+
+nf-core requires a Nextflow version 22.10.1 or higher, so you have to intall a more recent version first.
+
+For Nextflow versions newer than 22.10.1, it is necessary to mount the home directory using the command:
+
+```bash
+NXF_SINGULARITY_HOME_MOUNT=true
+```
+
+
+## Below are non-mandatory information
+
+>note:
+You will need access to the UzL OMICS cluster in order to run the pipeline. In doubt contact IT.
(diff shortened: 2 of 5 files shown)


conf/uzl_omics.config
```diff
@@ -0,0 +1,25 @@
...
+    max_time   = 72.h     
+}
+
+process {
+    executor = 'slurm'
+    queue = 'shortterm'
+    scratch = '$SCRATCH'
+}
+
+workDir = "$WORK/nf-core_rnaseq_workdir"
```
pontus, line 19: Forcing `workdir` in an institutional configuration does not make sense to me?

conf/uzl_omics.config
```diff
@@ -0,0 +1,25 @@
...
+
+singularity {
+    enabled = true
+    runOptions = "-B $WORK/nf-core_rnaseq_workdir -B $HOME/.nextflow/assets/nf-core/rnaseq/bin"
+    cacheDir = "$WORK/.singularity"
```
pontus, line 24: Mixing up the nextflow singularity cache with the directory used for (the different) cache used by singularity is probably not a good idea. Is there a problem with keeping it in the `workDir`?

conf/uzl_omics.config
```diff
@@ -0,0 +1,25 @@
...
+    executor = 'slurm'
+    queue = 'shortterm'
+    scratch = '$SCRATCH'
+}
+
+workDir = "$WORK/nf-core_rnaseq_workdir"
+
+singularity {
+    enabled = true
+    runOptions = "-B $WORK/nf-core_rnaseq_workdir -B $HOME/.nextflow/assets/nf-core/rnaseq/bin"
```
pontus, line 23: These are very specific for an institutional configuration.

My recommendation (for singularity) is to have the configuration bind all "project" file spaces by default, but for these, do you need more than `autoMounts = true`?

conf/uzl_omics.config
jfy133: Agreed, we have pipeline specific institutional comfigs for this! Check `conf/pipelines`

finjasygo (commented):
Updated the files according to your suggestions.

docs/uzl_omics.md
```diff
@@ -0,0 +1,32 @@
...
+The rnaseq nf-core pipeline has been successfully configured for use on the UzL OMICS cluster at the University of Luebeck.
+Implementation of nf-core pipelines for use on the UzL OMICS is in process. 
+
+To use, run the pipeline with `-profile uzl_omics`. This will download and launch the `uzl_omics.config` which has been pre-configured with a setup suitable for the UzL OMICS cluster. Using this profile, a docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.
+
+Before running the pipeline you will need to load Nextflow and Singularity using the environment module system on UzL OMICS cluster. You can do this by issuing the commands below:
+
+```bash
+module load nextflow
+module load singularity
```
pontus, line 12: Are loaded modules carried through to jobs or should this be in a `process.beforeScript` as well?

docs/uzl_omics.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+nf-core requires a Nextflow version 22.10.1 or higher, so you have to intall a more recent version first.
+
+For Nextflow versions newer than 23.07.0-edge, it is necessary to explicitly mount the home directory using the command:
```
pontus, line 17: There seems to be some buggy modules not being happy without a writable `$HOME`, is this because of this or some other thing?

docs/uzl_omics.md
```diff
@@ -0,0 +1,32 @@
...
+
+```bash
+NXF_SINGULARITY_HOME_MOUNT=true
+```
+
+You need to run the pipeline in `$WORK`. You can specify this in your batch-script, as in the following expample:
```
pontus, line 23: Is this really the case or an old remnant?

pontus:
Looks much better, but I made a few new comments on the documentation. 

Also, the linter still complains on the documentation page, can you please update? (Can e.g. be done with running `prettier` through docker if you don't want to install it on some system).

docs/uzl_omics.md
finjasygo: The modules are carried through. It's not needed in a `process.beforeScript`

docs/uzl_omics.md
finjasygo: You're right. In previous tests it failed that the pipeline didn't run in `$WORK`. But after testing it again, it works now. I erased it.

finjasygo (commented):
Fix according to the suggestions. Updated the documentation after running `prettier`.

docs/uzl_omics.md
pontus: It seems I made the last comments there ;)

Was the outcome that the singularity setup was intentionally broken (not allowing mount controls)? I guess that would limit to `$HOME` if that was the case, but then again, I'd definitely try raising it for a chance to have it fixed first.

finjasygo: Could you please explain what you mean and how we can get it fixed?

The singularity is installed by the HPC maintainers, and we are currently unsure why this problem occurs when we do not explicitly ask to mount `$HOME` using the global variable `NXF_SINGULARITY_HOME_MOUNT`.

In our understanding, it is rather a Nextflow related issue, because previous versions (e.g. 22.10.1) of Nextflow worked fine without this variable decoration.  Starting from version 23.07.0-edge, nextflow documentation says that the default is set to false when it comes to mounting the home directory.

pontus: As was mentioned in the slack thread, it is very likely a good idea to reach out to your systems people and let them now that their singularity configuration (disallowing user binds) makes things more difficulty for you.

I obviously don't know how your systems are set up exactly, and maybe everything is under users' home directories - then making sure nextflow doesn't disable home binds would be good enough (although the user experience is lousy as it currently requires setting that environment variable). But obviously that doesn't cover anything run outside of the home directories.

A setup that doesn't have singularity stopping if a bind is provided could use `singularity.autoMounts` to pass the binds nextflow knows of that would be required (although in some cases with links going back and forth, it might be a good idea to help with specifying some binds explicitly).

So, to sum up, I'd say the singularity configuration is weird if it doesn't allow the user specificying bind mounts (which it seemed from the slack thread), and I'd generally suggest trying to fix that since it breaks a lot of use cases.

finjasygo: We will work on fixing this issue. For now I have edited the documentation file and erased this part, to not cause any conflicts.

finjasygo (commented):
edited the documentation file

pontus (approved):

PR #616, opened 2024-01-19
Update scw.config to define charge account

dgrfs:
---
name: New Config
about: A new cluster config
---

### DIFF
--- a/conf/scw.config
+++ b/conf/scw.config
@@ -17,6 +17,11 @@ params {
   max_cpus = 20
   max_time = 72.h
 }
+
+// uncomment clusterOptions to add non-default slurm account
+// the slurm account is used to define which project to charge the CPU usage to. Usually 'scwXXXX'.
+
 process {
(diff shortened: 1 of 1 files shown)


jfy133 (changes requested):
This change won't work as users never come in contact with the config file: it is automagically (down)loaded by them by the pipeline.

I would rather suggest that the account name gets sets into a standard bash ENV variable, which then can be loaded here.

See CSIRO for a good example: https://github.com/nf-core/configs/blob/e8e642abfc0f3343c491aac1b7bcc268e0da9518/conf/csiro_petrichor.config#L23

jfy133:
@dgrfs do you still plan to to add this to your config? Do you need any further help?

If not, I would like to close it next week (we are doing a bit  of spring cleaning)

jfy133:
Ah I just saw you had this instead: https://github.com/nf-core/configs/pull/671

Will close this in favour of that as I presume that's your preferred method.

PR #621, opened 2024-01-30
Check if task.attempt isn't empty

nvnieuwk:
This change checks if `task.attempt` is empty `[:]`. If so, it defaults to `1`.

This to fix this error from popping up when a process fails before it has even started:
```
groovy.lang.MissingMethodException: No signature of method: static java.lang.Math.pow() is applicable fo
r argument types: (Integer, ConfigObject) values: [2, [:]]
```


matthdsm (approved):
Nice catch

PR #629, opened 2024-02-09
Update vsc_ugent

berombau:
Adds more support and docs for Singularity and GPU clusters.

@nvnieuwk

### DIFF
--- a/conf/vsc_ugent.config
+++ b/conf/vsc_ugent.config
@@ -33,20 +33,17 @@ process {
     stageInMode = "symlink"
     stageOutMode = "rsync"
     errorStrategy = { sleep(Math.pow(2, task.attempt ?: 1) * 200 as long); return 'retry' }
+    maxRetries    = 5
     // add GPU support with GPU label 
     // Adapted from https://github.com/nf-core/configs/blob/76970da5d4d7eadd8354ef5c5af2758ce187d6bc/conf/leicester.config#L26
     // More info on GPU SLURM options: https://hpc.vub.be/docs/job-submission/gpu-job-types/#gpu-job-types
     withLabel: use_gpu {
         // works on all GPU clusters of Tier 1 and Tier 2
         beforeScript = 'module load cuDNN/8.4.1.50-CUDA-11.7.0'
-        // TODO: Support multi-GPU configurations with e.g. ${task.ext.gpus}
-        clusterOptions = {"--gpus=1 --account=$tier1_project"}
+        // TODO: Support multi-GPU configuations with e.g. ${task.ext.gpus}
+        // only add account if present
+        clusterOptions = {"--gpus=1" + (tier1_project ? " --account=$tier1_project" : "")}
         containerOptions = {
-                // --containall or --no-home can break e.g. downloading big models to ~/.cache
-                // solutions to error 'no disk space left':
-                //   1. increase the memory of the job.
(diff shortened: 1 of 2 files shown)


conf/vsc_ugent.config
```diff
@@ -12,13 +6,10 @@ try {
     System.err.println("WARNING: Could not run sinfo to determine current cluster, defaulting to doduo")
 }
 
-if(!params.tier1_project && hostname.contains("dodrio")){
-    System.err.println("Please specify your project with --tier1_project in your Nextflow command or with params.tier1_project in your config file.")
-    System.exit(1)
-}
+def tier1_project = System.getenv("SBATCH_ACCOUNT") ?: System.getenv("SLURM_ACCOUNT")
```
nvnieuwk, line 9: Both of these don't seem to contain the project name for our starting grant. Maybe we should return to the previous way with the params but you can set those environment variables as defaults instead

conf/vsc_ugent.config
```diff
@@ -42,7 +33,25 @@ process {
     stageInMode = "symlink"
     stageOutMode = "rsync"
     errorStrategy = { sleep(Math.pow(2, task.attempt ?: 1) * 200 as long); return 'retry' }
-    maxRetries    = 5
```
nvnieuwk, line 45: Please don't remove this as it's a failsafe for inconsistencies from the scheduler. Maybe lower this to 3 as a compromise? You can always turn it off in your own configs if you don't want that

conf/vsc_ugent.config
```diff
@@ -42,7 +33,25 @@ process {
...
+    // add GPU support with GPU label 
+    // Adapted from https://github.com/nf-core/configs/blob/76970da5d4d7eadd8354ef5c5af2758ce187d6bc/conf/leicester.config#L26
+    // More info on GPU SLURM options: https://hpc.vub.be/docs/job-submission/gpu-job-types/#gpu-job-types
+    withLabel: use_gpu {
+        // works on all GPU clusters of Tier 1 and Tier 2
+        beforeScript = 'module load cuDNN/8.4.1.50-CUDA-11.7.0'
+        // TODO: Support multi-GPU configurations with e.g. ${task.ext.gpus}
+        clusterOptions = {"--gpus=1 --account=$tier1_project"}
```
nvnieuwk, line 43: Can you only add the account flag here if the `tier1_project` variable is set? I think this will cause issues for tier 2 clusters otherwise

nvnieuwk (changes requested):
Some structural changes that need to happen. I'll have a look at the doc changes when these are fixed

conf/vsc_ugent.config
berombau: These environment variables SLURM users have to define themselves (like I added in the docs and e.g [here](https://docs.dkrz.de/blog/2017/how-to-set-the-default-slurm-project-account.html)). This is the [official way](https://slurm.schedmd.com/sbatch.html#SECTION_INPUT-ENVIRONMENT-VARIABLES) to pass the --account option to SLURM, there is no need for a --tier1_project option.

nvnieuwk (commented):
LGTM, can you fix the issues that popped up in the tests first? I'll approve if those are fixed

nvnieuwk (approved):
LGTM!!

PR #651, opened 2024-03-16
Set process medium to use genomics/normal queue

edmundmiller:
Should use the genomics queue. The 6 cpus and 36GB of memory is just plain off.


alyssa-ab (approved):
lgtm

PR #666, opened 2024-04-15
New Config: Novo Nordisk's Einstein cluster

ashotmarg:
---
name: New Config
about: A new cluster config
---

### DIFF
--- a/conf/einstein.config
+++ b/conf/einstein.config
@@ -1,0 +1,31 @@
+/*
+* ------------------------------------------------------------------
+*   Config file for nextflow pipelines for Novo Nordisk's Einstein cluster
+* ------------------------------------------------------------------
+*/
+
+params {
+    config_profile_description = "Novo Nordisk's Einstein cluster profile"
+    config_profile_contact = 'Ashot Margaryan (aoqn@novonordisk.com)'
+    max_memory = "750.GB"
+    max_cpus = 64
+    max_time = "240.h"
+}
+
+process {
+    executor = 'slurm'
+    scratch = true
+    maxRetries = 3
+    scratch = "/scratch/users/${USER}"
+}
+
+executor {
+    queueSize = 50
+    submitRateLimit = "10/1sec"
+}
+
+singularity {
+    enabled = true
+    autoMounts = true
+    cacheDir = "/scratch/users/${USER}"
+}
--- a/docs/einstein.md
+++ b/docs/einstein.md
@@ -1,0 +1,22 @@
+# nf-core/configs: Novo Nordisk's Einstein Configuration
+
+## Using the Einstein config profile
+
+Before running any pipeline in Novo Nordisk's Einstein cluster, `Nextflow` will need to be installed in the conda environment being used.
+
+To use, run the pipeline with `-profile einstein` (one hyphen).
+This will download and launch the [`einstein.config`](../conf/einstein.config) which has been pre-configured with a setup suitable for the einstein server.
+It will enable `Nextflow` to manage the pipeline jobs via the `Slurm` job scheduler.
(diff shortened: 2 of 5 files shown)


jfy133:
@nf-core-bot fix linting

conf/einstein.config
```diff
@@ -0,0 +1,31 @@
...
+    max_memory = "750.GB"
+    max_cpus = 64
+    max_time = "240.h"
+}
+
+process {
+    executor = 'slurm'
+    scratch = true
+    maxRetries = 3
+    scratch = "/scratch/users/${USER}"
```
jfy133, line 19: Has this been tested? It may fail as the `${USER}` will not resolve witthin the config

jfy133 (commented):
Basdically my comments are: fine if it's tested and works as expectd :grimacing:

conf/einstein.config
ashotmarg: Users have access to only their own scratch directory at the moment, so that's why I want to keep the ${USER} or the equivalent option in the scratch path.

docs/einstein.md
```diff
@@ -0,0 +1,22 @@
...
+Using this profile, `Docker` image(s) containing required software(s) will be downloaded, and converted to `Singularity` image(s) if needed before execution of the pipeline.
+
+
+## Below are non-mandatory information on e.g., required modules
+
+Before running the pipeline you will need to have a working Nextflow installation and load Singularity using the environment module system on the Einstein cluster. You can do this by doing for example:
+
+```bash
+# Load Singularity environment modules
+module load singularity/3.8.1
```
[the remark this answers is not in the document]
ashotmarg: I actually had this already in the beforeScript part of this config as comment as I wasn't sure about this, but then removed it, as I didn't want to "force" any specific version of the module and wanted to keep it more general. Also if I include it here, I should probably also have to use `module purge` to remove already loaded singularity versions.

conf/einstein.config
jfy133: > Thanks for the prompt look James! This works, as I guess the $USER is a unix variable. This can of course be overridden in some cases, so do you mean I should put something like `whoami` or `logname` instead?

Nope, I meant more that in some cases  groovy evaluates it as a literal string or it failed. But if it works here, all good :+1:

docs/einstein.md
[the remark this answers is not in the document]
jfy133: I only mention this as in some clusters the compute node the job goes to would not  have singularity loaded and will crash. I'm assuming  here yo uare expecting uses to export their environment with SBATCH submissions or something?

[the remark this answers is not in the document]
ashotmarg: Actually, you are probably right, I will add the `module purge; module load singularity/3.8.1` part to the `beforeScript`, and will also change the ${USER} with `whoami`, I just tested and it worked, and is probably better, as the $USER can be overwritten as a variable.
Thanks for the helpful comments.

conf/einstein.config
```diff
@@ -0,0 +1,32 @@
...
+
+executor {
+    queueSize = 50
+    submitRateLimit = "10/1sec"
+}
+
+singularity {
+    enabled = true
+    autoMounts = true
+    cacheDir = "/scratch/users/${USER}"
```
jfy133, line 31: `whoami` here too?

ashotmarg:
Hi again @jfy133 , could you approve it now?

jfy133 (approved):
See last comment otherwise LGTM

conf/einstein.config
ashotmarg: nice catch!

PR #668, opened 2024-04-15
Changed the aoqn, which didn't work, with aoqn

ashotmarg:
---
name: Correct config file
about: Correction in the existing config
---

### DIFF
--- a/conf/einstein.config
+++ b/conf/einstein.config
@@ -17,7 +17,7 @@ process {
     beforeScript = 'module purge; module load singularity/3.8.1'
     scratch = true
     maxRetries = 3
-    scratch = "/scratch/users/`whoami`"
+    scratch = "/scratch/users/${USER}"
 }
 
 executor {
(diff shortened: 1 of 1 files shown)


ashotmarg:
Hi @jfy133 , sorry, I have to bug you with this again. For some weird reason the second '`whoami`' command substitution doesn't actually work, i.e., i get the literal '`whoami`'... So, I guess I either had to create extra groovy variable to extract the username and pass to the cacheDir, or just use the ${USER}, which is working.

mashehu (approved):

jfy133:
I think in this case it's safer to use something like @nvnieuwk does at Ghent: https://github.com/nf-core/configs/blob/9984beb43d1f2889b2238066b83e2ec4fc908a04/conf/vsc_ugent.config#L9

Maybe there is a system function to get the output of commands so you can run `whoami`?

ashotmarg:
Yeah, there seem to be few other ways of doing this in groovy, but perhaps in this case just the plain ${USER} is more straightforward.

PR #669, opened 2024-04-17
Crick config - Update max resources for ncpu default queue

chris-cheshire:
---
name: New Config
about: A new cluster config
---

### DIFF
--- a/conf/crick.config
+++ b/conf/crick.config
@@ -1,6 +1,6 @@
 //Profile config names for nf-core/configs
 params {
-    config_profile_description = 'The Francis Crick Institute CAMP HPC cluster profile provided by nf-core/configs.'
+    config_profile_description = 'The Francis Crick Institute NEMO HPC cluster profile provided by nf-core/configs.'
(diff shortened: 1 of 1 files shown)


conf/crick.config
```diff
@@ -1,6 +1,6 @@
 //Profile config names for nf-core/configs
 params {
-    config_profile_description = 'The Francis Crick Institute CAMP HPC cluster profile provided by nf-core/configs.'
+    config_profile_description = 'The Francis Crick Institute NEMO HPC cluster profile provided by nf-core/configs.'
```
jfy133, line 3: I guess too late, but given it has a distinct name: do you have multiple? If so maybe it makes sense to have one config per cluster e.g. `--profile nemo`

jfy133 (approved):

conf/crick.config
chris-cheshire: not too late! But no, the cluster name got changed to nemo from camp, camp is gone

PR #674, opened 2024-04-24
Add dummy configs for all missing configs

jfy133:
This is for allowing us to uncomment out the part in the pipeline template.

I wonder if it just works 'as is' or if we have to add an explicit profile or not... ideas @maxulysse ?

### DIFF
--- a/.github/workflows/update-pipeline-configs.yml
+++ b/.github/workflows/update-pipeline-configs.yml
@@ -1,0 +1,78 @@
+name: Update Pipeline Configs
+on:
+  schedule:
+    - cron: '0 0 * * *' # Run every day at midnight
+  workflow_dispatch:
+
+
+jobs:
+  update_configs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Fetch pipeline names
+        run: |
+          curl -o pipeline_names.json https://nf-co.re/pipeline_names.json
+
+      - name: add test pipeline to pipeline_names.json #TODO remove this testing step after successful tests
+        run: |
+          jq '.pipeline += ["test"]' pipeline_names.json > tmp.json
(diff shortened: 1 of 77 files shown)


maxulysse (approved):
LGTM, we just need to be sure we add any new pipeline

mashehu:
One thing I find sub-optimal: we now have `conf/pipeline/` and `pipeline/`. too easy to confuse the two, imo

maxulysse:
> One thing I find sub-optimal: we now have `conf/pipeline/` and `pipeline/`. too easy to confuse the two, imo

both have different usage, but I see your point

mashehu:
yes, different usage and same name is never good

jfy133:
> If you want I can write a github action which periodically checks https://nf-co.re/pipeline_names.json and creates that.

That would be _awesome_

jfy133:
Should I wait for the GHA @mashehu before merging? But otherwise I will prepare the tools template to uncomment the relevant code, and update our new-project-guidance

mashehu:
yes, let's add the GHA to this PR. Any chance we can use a different name than `pipeline/` for the placeholders?

jfy133:
> yes, I will add the GHA to this PR. Any chance we can use a different name than `pipeline/` for the placeholders?

No I don't think so. It's already embedded in existing pipelines for a while, so if we changed it would break compatibility with old releases. @maxulysse correct me if I'm wrong

mashehu:
okay, GHA is there. as usual need to really test it in production (I ran it locally with `act` and it looked fine, but we will see). Left some testing code in it, so we will need a clean-up PR after this one is merged and the workflow has been tested

jfy133:
Looks good to me @mashehu ! 

If you can both review: https://github.com/nf-core/tools/pull/2936/ and re-review this one, then i will merge this PR and shortly after the tools one

mashehu (approved):

Issue #687, opened 2024-05-18
azurebatch config doesn't expose tokenDuration parameter, causing sadness


tanaes:
I recently discovered the existence of the `azure.storage.tokenDuration` parameter in the wrong way -- that is, by having a large and expensive workflow fail spectacularly 2/3 of way through compute due to token expiration. 🙃 The nature of this failure appears to be a particular bummer, as it seems to have interfered with cacheing, requiring me to bodge together a fix to prevent needless re-execution of most of the expensive steps. 

I'd propose exposing this parameter in the configuration file!

maxulysse:
this should be covered by #688

PR #688, opened 2024-05-18
Azurebatch: expose tokenDuration parameter

tanaes:
Addresses issue #687


conf/azurebatch.config
```diff
@@ -8,6 +8,7 @@ params {
     storage_name      = null
     storage_key       = null
     storage_sas       = null
+    storage_duration  = null
```
mashehu, line 11:
```suggestion
    storage_duration  = '48h'
```
Based on the default value given below, but no azure user myself.

adamrtalbot (approved):
Bumped the default to the same as Nextflow.

jfy133 (approved):

jfy133 (commented):
Not a blocker, but I wonder if such extra parameters will result in validation warnings in nf-core pipelines? (Unknown parama type things)

maxulysse (approved):

PR #696, opened 2024-05-24
Patch biowulf config

qiyubio:
---
name: Update the cluster config
about: patches
---


jfy133 (approved):

conf/biowulf.config
```diff
@@ -6,48 +6,41 @@ params {
...
     autoMounts = true
-    cacheDir = "/data/$USER/singularity"
-    envWhitelist='https_proxy,http_proxy,ftp_proxy,DISPLAY,SLURM_JOBID,SINGULARITY_BINDPATH'
+    cacheDir = "/data/$USER/nxf_singularity_cache"
+    envWhitelist='https_proxy,http_proxy,ftp_proxy,DISPLAY,SLURM_JOB_ID,SINGULARITY_BINDPATH'
 }
 
 env {
-    SINGULARITY_CACHEDIR="/data/$USER/singularity"
+    SINGULARITY_CACHEDIR="/data/$USER/.singularity"
```
jfy133, line 21: Just noticed this, shouldn't this be the same as the path above?

conf/biowulf.config
```diff
@@ -3,51 +3,44 @@ params {
...
```
edmundmiller, line 18:
```suggestion
    SINGULARITY_CACHEDIR="/data/$USER/nxf_singularity_cache"
```

edmundmiller (dismissed):
I have only seen them ever be the same. 

So I think, make them match or reveal your secrets 😁

qiyubio:
There is bug in this setting, the time without .h will cause user's slurm job submitted with unlimited time, and the job will never got started since it excessed the wall time of slurm partition. Could my pull request got approved? Which has the newest resource-limits, and SINGULARITY_CACHEDIR is removed. 

 `   resourceLimits = [
        memory: 224,
        cpus: 32,
        time: 72
    ]
`

PR #698, opened 2024-05-27
Remove igenome parameters from genotoul config.

noirot:
---
name: Update genotoul Config
about: Remove igenome parameters 
---


maxulysse:
so false by default?
How do you deal with references on genotoul then?

noirot:
Sorry I thought it generate an error, in fact it's just a warning. 
So I can keep like this

PR #707, opened 2024-06-14
Janelia queue size

krokicki:
We'd like to update Janelia's configuration so that there is a default queueSize of 500. We often need to run many Dask workers and the 200 default is not enough. I'm also making this a parameter in case we need to increase it further for special cases.


jfy133 (approved):

krokicki:
Thanks a lot, James!

PR #713, opened 2024-07-03
TWEAK: Add ENA prep profile to nf-core/eager pipeline config for EVA

TCLamnidis:
add ENA prep profile

This profile simply skips any skippable steps, and runs AR without any length/quality filtering and without merging reads.


jfy133 (approved):
Good enough for now :)

PR #716, opened 2024-07-16
Remove module charliecloud/0.30

nschan:
Charliecloud@0.30 is incompatible with the strategy nextflow uses to create temporary containers.

There is no recent version of charliecloud available as a default module, so this cannot be bumped. 
I would remove this and rely on users to load a recent charliecloud version.

### DIFF
--- a/conf/biohpc_gen.config
+++ b/conf/biohpc_gen.config
@@ -8,7 +8,6 @@ params {
 process {
     executor = 'slurm'
     queue    = { task.memory <= 1536.GB ? (task.time > 2.d || task.memory > 384.GB ? 'biohpc_gen_production' : 'biohpc_gen_normal') : 'biohpc_gen_highmem' }
-    module   = 'charliecloud/0.30'
     clusterOptions = '--clusters=biohpc_gen'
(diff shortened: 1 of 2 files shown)


jfy133:
Sounds reasonable to me but 8 would suggest documenting this in the readme, ideally with which ancient version works

nschan:
That is a very good suggestion. However, I do not remember which combination of ancient versions work, and even less which combination required which specific work-around to make it run. There were multiple issues with charliecloud and nextflow for a while, which have been resolved as of 24.04.2. 
Instead of providing workarounds to run ancient versions, I opted for explaining how to use recent versions on this cluster.
Convenience of this whole attempt to run nf-core on this cluster (and other clusters that use charliecloud as a container engine with e.g. SLURM) anyway hinges on nf-core/tools/pull/2937 getting into master.

jfy133 (approved):
LGTM as an outsider!

jfy133:
@nf-core-bot fix linting

jfy133:
(if nf-core-bot is here... I can never remember)

PR #725, opened 2024-08-07
Update vsc_calcua

pmoris:
Some bugfixes and improvements for the CalcUA VSC config.

- Bug fix for unknown config attribute error.
- Clearer warnings about singularity/apptainer and nextflow tmp and cache dirs.
- Improvements to checks whether process is running on a correct partition versus login node, and with the appropriate resources.


pmoris:
- Fixed automatic memory allocation. I didn't realize there were multiple possible SLURM environment variables that could keep track of the available amount of memory (`SLURM_MEM_PER_NODE` vs `SLURM_MEM_PER_CPU`).

conf/vsc_calcua.config
```diff
@@ -31,6 +44,16 @@ if ( host == "antwerpen" ) {
             }
         }
     }
+    // APPTAINER_CACHEDIR/SINGULARITY_CACHEDIR
+    if ( !apptainer_cachedir ) {
+        System.err.println("\nWARNING: APPTAINER_CACHEDIR/SINGULARITY_CACHEDIR environment variable was not found.\nPlease add the line 'export APPTAINER_CACHEDIR=\"\${VSC_SCRATCH}/apptainer/cache\"' to your ~/.bashrc file (or set it with sbatch or in your job script).\nUsing the default storage location of Singularity/Apptainer ~/.apptainer/cache/. Read more about why this should be avoided in the VSC docs: https://docs.vscentrum.be/software/singularity.html#building-on-vsc-infrastructure\n")
```
jfy133, line 49: System.err vs warning, maybe should say 'ERROR' rather than warning? I would find that confusing (unless the `System.err` doesn't actually error/stop hte pipeline.

jfy133 (approved):
OK this is a ultra-mega-complicated-config :sweat_smile: , but assuming it's been tested, I don't see anything majorly flawed :+1:

pmoris:
> OK this is a ultra-mega-complicated-config 😅 , but assuming it's been tested, I don't see anything majorly flawed 👍

I might have gone a little overboard with this one... 

The basic premises are:

1. Allow the choice between SLURM scheduled jobs and local execution on a single worker node (`*_slurm` vs `*_local`).
    -  The former has more overhead and waiting time between tasks, but also allows more parallel tasks. The latter is faster, but limited to the CPU/RAM of a single node.
    - Since each of our SLURM partitions contains a different type of nodes (CPU, memory, wall time, etc.), I separated these out into sub-profiles (`zen2_*`, `skylake_*`, etc.). A single profile would have `max_*` values that could be too high (or too restrictive) depending on the partition a job is submitted to, wouldn't it?
2. Dynamically set `max_*` values for local execution profiles, by setting these based on the number of requested resources as reported by SLURM (and default to the max of the partition/node type otherwise).
3. Add several checks and warnings:
    - Check if singularity cache, tmp and nextflow cache are set properly. Our HPC has strict file quotas and low storage worker nodes that could be exceeded/filled otherwise.
    - Avoid runs on logins nodes.
4. Avoid crashing the GitHub actions by skipping various parts of the config if the hostname does not match that of our HPC. 
5. Lots of documentation on how to install/run nextflow, things to avoid, what the default directories are, overview of the different resources and matching profiles, etc.

I'm open to any and all suggestions to improve this config though, especially if they can simplify this behemoth of a config!

I just noticed that the [vsc_ugent.config](https://github.com/nf-core/configs/blob/master/conf/vsc_ugent.config) recently switched to dynamically determining the current SLURM partition, which eliminates the needs for different subprofiles. So I might put that on my todo list. I'd still need two subprofiles to specify SLURM vs local executor, but if one of them is set as the default, that'd still simplify things.

conf/vsc_calcua.config
pmoris: Thanks for having a look!

System.err doesn't actually stop the pipeline - I believe it simply prints to stder instead of stdout - but this did confuse me as well coming back to the config after not looking at it for a while. 

The only places the config actually halts the process is where there's an explicit `System.exit(1)` statement (e.g., when mis-matching profiles/partitions), or a `try catch` (nxf cache dir not creatable,). I've tried to restrict these to bigger issues, and just warn the users for less severe sub-optimal usage (e.g., not setting a singularity cache).

I'll change those to ERROR instead of warning to make the distinction more clear.

conf/vsc_calcua.config
```diff
@@ -31,6 +44,16 @@ if ( host == "antwerpen" ) {
...
+        // TODO: optional exit out here.
```
nvnieuwk, line 50: I would make a pipeline exit here to enforce user to set these variables. Better to have this than having their home dirs fill up, which is a lot harder to debug :grin:

jfy133:
> > OK this is a ultra-mega-complicated-config 😅 , but assuming it's been tested, I don't see anything majorly flawed 👍
> 
> I might have gone a little overboard with this one...
> 
> The basic premises are:
> 
>     1. Allow the choice between SLURM scheduled jobs and local execution on a single worker node (`*_slurm` vs `*_local`).
>        
>        * The former has more overhead and waiting time between tasks, but also allows more parallel tasks. The latter is faster, but limited to the CPU/RAM of a single node.
>        * Since each of our SLURM partitions contains a different type of nodes (CPU, memory, wall time, etc.), I separated these out into sub-profiles (`zen2_*`, `skylake_*`, etc.). A single profile would have `max_*` values that could be too high (or too restrictive) depending on the partition a job is submitted to, wouldn't it?
> 
>     2. Dynamically set `max_*` values for local execution profiles, by setting these based on the number of requested resources as reported by SLURM (and default to the max of the partition/node type otherwise).
> 
>     3. Add several checks and warnings:
>        
>        * Check if singularity cache, tmp and nextflow cache are set properly. Our HPC has strict file quotas and low storage worker nodes that could be exceeded/filled otherwise.
>        * Avoid runs on logins nodes.
> 
>     4. Avoid crashing the GitHub actions by skipping various parts of the config if the hostname does not match that of our HPC.
> 
>     5. Lots of documentation on how to install/run nextflow, things to avoid, what the default directories are, overview of the different resources and matching profiles, etc.
> 
> 
> I'm open to any and all suggestions to improve this config though, especially if they can simplify this behemoth of a config!
> 
> I just noticed that the [vsc_ugent.config](https://github.com/nf-core/configs/blob/master/conf/vsc_ugent.config) recently switched to dynamically determining the current SLURM partition, which eliminates the needs for different subprofiles. So I might put that on my todo list. I'd still need two subprofiles to specify SLURM vs local executor, but if one of them is set as the default, that'd still simplify things.

I don't have time to read through fully currently (in the middle of running/teaching a summer school), so I would say you can merge in otherwise, depending on @nvnieuwk 's comment :)

conf/vsc_calcua.config
pmoris: Done!

pmoris:
Incorporated your suggestions and tried to improve the docs/comments a bit more. Ready for merging! Thanks for the help!

PR #728, opened 2024-08-12
Patch binac config

fbartusch:
---
name: Update the cluster config
about: patches
---

The way  how we provide Apptainer/Singularity on the cluster changed. We don't have environment modules any more, but Apptainer is installed as OS package on each compute node.

I also added myself (one of the BinAC admins) as the profile contact, as Alexander Peltzer doesn't use the BinAC anymore.

### DIFF
--- a/conf/binac.config
+++ b/conf/binac.config
@@ -1,17 +1,16 @@
 //Profile config names for nf-core/configs
 params {
     config_profile_description = 'BINAC cluster profile provided by nf-core/configs.'
-    config_profile_contact = 'Alexander Peltzer (@apeltzer)'
-    config_profile_url = 'https://www.bwhpc-c5.de/wiki/index.php/Category:BwForCluster_BinAC'
+    config_profile_contact = 'Felix Bartusch (@fbartusch)'
+    config_profile_url = 'https://wiki.bwhpc.de/e/BinAC'
 }
 
 singularity {
     enabled = true
-    envWhitelist = 'TZ'
+    cacheDir = '/beegfs/work/container/apptainer_cache'
 }
 
 process {
-    beforeScript = 'module load devel/singularity/3.4.2'
     executor = 'pbs'
     queue = { task.memory >= 128.GB ? 'smp': task.time <= 20.m ? 'tiny' : task.time > 48.h ? 'long' : 'short'}
 }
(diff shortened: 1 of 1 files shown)


apeltzer (approved):
Should be fine, thanks for taking over @fbartusch 👍🏻

conf/binac.config
```diff
@@ -1,17 +1,17 @@
...
+    config_profile_name = 'binac'
+    config_profile_description = 'BinAC cluster profile provided by nf-core/configs.'
+    config_profile_contact = 'Felix Bartusch (@fbartusch)'
+    config_profile_url = 'https://wiki.bwhpc.de/e/BinAC'
 }
 
 singularity {
     enabled = true
-    envWhitelist = 'TZ'
+    cacheDir = '/beegfs/work/container/apptainer_cache'
```
pontus, line 11: Just to check, is this a globally writable directory (possibly with some sense applied, e.g. sticky bit set)? Otherwise it might make more sense to use `libraryDir` (which for some reason doesn't show up in the config documentation page for me now but is documented in https://www.nextflow.io/docs/latest/container.html#singularity).

fbartusch: @pontus Thans for the suggestions. Yes, this is a globally writable directory. The intention is to store each container only once on the file system. Our users often forget to clean up their data (including containers) when they leave the university, and space on the file system is limited.

I’ve noticed that in other configurations, each user has their own cacheDir. Thinking about it... with one shared directory for all users, a malicious user could store manipulated containers. With that in mind, I’m inclined to change the configuration and create a directory for each user.

What is the intention behind `libraryDir`? It seems to function as another directory where Nextflow searches for images. Is `libraryDir` essentially a secondary cache that is checked first?

pontus: Yes, `libraryDir` allows for having another cache that is checked first (but where writes aren't tried).

The purpose it fills is to allow for a central cache where administrators or trusted users (or a script caching images used by various common pipelines) can manage container images without the restrictions on writing causing problems for users that aren't able to write their.

(Using a cache is a very good thing as it cuts down on duplication and gets rid of unneeded downloads. But allowing everyone to manage it isn't very good from a security standpoint.)

fbartusch: Thanks for the clarification regarding `libraryDir`. This seems like a good option, as I currently know exactly which nf-core pipelines and revisions my users want to use. I will test a configuration with `libraryDir` set and then update the configuration to also set a different `cacheDir` for each user.

PR #730, opened 2024-08-13
Update vsc_calcua - simplify all the things

pmoris:
Based on the feedback in the [previous PR](https://github.com/nf-core/configs/pull/725) (thanks @jfy133 and @nvnieuwk ) and inspired by the config for the UGent VSC, I've tried to simplify the CalcUA config.

See the commit messages for a full breakdown, but briefly:

- Use a sane default profile that uses slurm scheduling.
- Move single node execution mode to a sub-profile.
- Dynamically set the queue and resource limits based on the partition to which the job was submitted.
- Remove dynamic queue selector based on task requirements because it was a pain to debug.
- Increase the queue size and lower the timeout setting.

### DIFF
--- a/conf/vsc_calcua.config
+++ b/conf/vsc_calcua.config
@@ -59,28 +59,6 @@ if ( System.getenv("VSC_INSTITUTE") == "antwerpen" ) {
     }
 }
 
-// Function to check if the selected partition profile matches the partition on which the head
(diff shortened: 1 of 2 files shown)


nvnieuwk (approved):
LGTM! Very nice!

PR #732, opened 2024-08-14
Update binac.config

fbartusch:
---
name: binac config
about: update binac config
---

Sorry for the second pull request within a week. As @pontus  pointed out in the last (already merged) /pull/728 request, setting a single global `cacheDir` is not ideal for security. I also added a `libraryDir`.

### DIFF
--- a/conf/binac.config
+++ b/conf/binac.config
@@ -8,7 +8,8 @@ params {
 
 singularity {
     enabled = true
-    cacheDir = '/beegfs/work/container/apptainer_cache'
+    libraryDir = '/beegfs/work/container/apptainer_library'
+    cacheDir = '/beegfs/work/container/apptainer_cache/$USER'
 }
 
 process {


conf/binac.config
```diff
@@ -8,7 +8,8 @@ params {
 
 singularity {
     enabled = true
-    cacheDir = '/beegfs/work/container/apptainer_cache'
+    libraryDir = '/beegfs/work/container/apptainer_library'
+    cacheDir = '/beegfs/work/container/apptainer_cache/$USER'
```
pontus, line 12: Just to check, have you tested this? My recollection is that nextflow isn't happy if you point it at an non-existing cache directory (and while it shouldn't be super hard, if it gets into setting up a system to create these, maybe it's not worth caring about if you knew what pipelines people would run and could pull those images into the library directory anyway).

fbartusch: Nextflow creates a non-existing cache directory. `/beegfs/work/container/apptainer_cache/` still exists and Nextflow creates the `$USER`part. I tested it by deleting my directory in the cache and running a pipeline from scratch. The pipeline recreates the directory.
But I made a big mistake by using single quotes, `$USER` doesn't get resolved and it creates a directory called `$USER`. A very embarrassing mistake.

PR #745, opened 2024-09-06
Add BlueBEAR cluster config

alexlyttle:
---
name: BlueBEAR
about: A new cluster config
---


conf/bluebear.config
```diff
@@ -0,0 +1,40 @@
...
+    // Maximum resources available
+    max_memory = 512.GB
+    max_cpus = 112
+    max_time = 10.d
+
+    // Schema validation
+    validationSchemaIgnoreParams = 'genomes,igenomes_base,project'
+
+    // Cluster-specific params
+    project = slurm_job_account
```
jfy133, line 18: Having non-pipeline parameters here may fail some (and increasingly more pipelines).

A better approach is to tell users to set a bash environment variable with the information needed, and yse that information (in this case down in your clusterOptions).

For example, see https://github.com/nf-core/configs/blob/2900904b814873e911218c60972c15db7e119341/conf/pawsey_setonix.config#L26

docs/bluebear.md
```diff
@@ -0,0 +1,181 @@
...
+module load bear-apps/2022b
+module load Nextflow/24.04.2
+```
+
+## Apptainer/Singularity
+
+BlueBEAR comes with Apptainer installed for running processes inside containers. Our configuration makes use of this
+when executing processes.
+
+We advise you create a directory in which to cache images using the following environment variable. For example,
```
jfy133, line 19: Maybe mention it's a good idea to put the `export` command in a `.bashrc` or similar?

docs/bluebear.md
```diff
@@ -0,0 +1,181 @@
...
+export NXF_SINGULARITY_CACHEDIR="/rds/projects/_initial_/_project_/.apptainer"
+
+nextflow run nf-core/rnaseq -profile bluebear,test --outdir test_results
+```
+
+## Clean work directory
+
+Nextflow caches pipeline files in a work directory (default is `work`). This is useful if you need to resume a job
+after a failure (using the `-resume` option). However, the `work` directory can fill up quickly. Regularly cleaning the
+work directory avoids filling up your project quota.
```
jfy133, line 82: You could consider by default adding the `cleanup = true` [^1] option to have the emptying of the `work/` happen by default, and to override this use the `debug` profile in addition to the institutional profile.

For examples: https://github.com/search?q=repo%3Anf-core%2Fconfigs%20cleanup%20%3D%20true&type=code

[^1]: Docs: https://www.nextflow.io/docs/latest/config.html#miscellaneous

jfy133 (commented):
A couple of minor comments, if you disagree we can proceed.

Note, so you don't ave to wait for approval for the tests to run - consider joining the nf-core github organisation, instructions here: https://nf-co.re/join

conf/bluebear.config
alexlyttle: Thanks, that sounds good. In fact `System.getenv("SLURM_JOB_ACCOUNT")` could be substituted there, since it is unlikely someone would want individual processes to be submitted under a different account.

docs/bluebear.md
alexlyttle: I was a little concerned about the warning in the Nextflow docs about this. It suggests `cleanup = true` prevents use of `-resume`. If cleanup only occurs on a successful run, is `-resume` still possible after a failed run?

jfy133: It only prevents `-resume` if the pipeline completes with exit code 0 (so no errors).

If the pipeline hits an error (exit code >1), all the files in `work` are retained and will not be deleted

The only case that it might be problematic is if someone is playing with parameters in later stages of a pipeline, and want to resume from midway but with a different parameter, yes in this case this `resume` would not work (if that makes sense)

alexlyttle: Ah fantastic, I misunderstood the docs slightly. This sounds like a good addition, thanks!

docs/bluebear.md
alexlyttle: Done!

jfy133 (approved):

jfy133:
Good to go @alexlyttle !

With your new found nf-core org membership, you may do the honours! (At least I think you should be able to merge yourself now 😅)

PR #746, opened 2024-09-09
roslin.config - Exclude Rocky Linux nodes

sguizard:
A small modification to exclude Rocky Linux 9 nodes that are not fully ready to run nextflow jobs.


pontus (approved):

sguizard:
@pontus Thanks!

PR #752, opened 2024-09-23
Format all config files for consistency

jfy133:
@bentsherman made me do it

In preparation for a build update to copy the old `max_*` parameters into the new Nextflow neative `resourceLimits` directive


conf/genouest.config
```diff
@@ -17,8 +17,9 @@ process {
 
 params {
     igenomes_ignore = true
-    igenomesIgnore = true //deprecated
-    max_memory = 3000.GB
-    max_cpus = 160
-    max_time = 336.h
+    igenomesIgnore  = true
+    //deprecated
```
mirpedrol, line 21:
```suggestion
    igenomesIgnore  = true //deprecated
```

conf/imperial.config
```diff
@@ -82,21 +85,21 @@ profiles {
 }
 
 executor {
-    $pbspro {
-        queueSize = 49
+    '$pbspro' {
```
mirpedrol, line 88: Is it ok to use single quotes? will the variable be replaced?

mirpedrol (approved):
Leaving two tiny comments, otherwise LGTM

davidecarlson:
The seawulf config LGTM.

wassimsalam01:
I've just opened a PR to fix mine, kindly check and approve please. Thank you

conf/genouest.config
jfy133: Will put it above

conf/imperial.config
jfy133: I think if it's single quotes it won't be interpreted, so I think it should be OK

nextflow.config
```diff
@@ -1,2 +1,2 @@
 params.custom_config_base = "."
-includeConfig("nfcore_custom.config")
+includeConfig ("nfcore_custom.config")
```
mirpedrol, line 2: I think the parenthesis are not needed here
```suggestion
includeConfig "nfcore_custom.config"
```

conf/imperial.config
jfy133: @bentsherman can you confirm it's OK to put in quotes?

jfy133:
@ggrimes your profile is the one I'm least sure about the change, but we will we can use a better work around if it breaks something (`System.getenv(TMPDIR)` like other shave used)

vsmalladi (approved):
lgtm

ggrimes:
> @ggrimes your profile is the one I'm least sure about the change, but we will we can use a better work around if it breaks something (`System.getenv(TMPDIR)` like other shave used)

how can I test this?

mirpedrol:
Merging this now, please let us know if anything fails on the pipelines end

conf/imperial.config
bentsherman: should be fine with single quotes or no quotes

jfy133: Thank you!

jfy133:
> > @ggrimes your profile is the one I'm least sure about the change, but we will we can use a better work around if it breaks something (`System.getenv(TMPDIR)` like other shave used)
> 
> how can I test this?

Run a pipelines than use the local and pbspro executors and check the settings defined in your config are applied correctly. That said Ben said it should be fine anyway :+1:

PR #759, opened 2024-09-24
WIP : Updated text plus preference of singularity profile

chriswyatt1:
Update to UCL Myriad config

Made all export env lines universal with $HOME
Requested users use `-profile singularity`, as apptainer does not work for some pipelines. 

@FernandoDuarteF , could you double check with works please. Would be good to quickly check this all works for a couple of pipelines, but should be fine now. Thanks

### DIFF
--- a/docs/ucl_myriad.md
+++ b/docs/ucl_myriad.md
@@ -10,8 +10,9 @@ Before running an nf-core pipeline you will need to set up the requirements and 
 
 You will need the java 11 or later to install and run Nextflow. Different java versions are available in the cluster. To avoid any compatibily issues, use the latest version of `java/temurin`. You can do this by placing the next line inside your `.bashrc`:
 
-```
+```bash
 module load java/temurin-17/17.0.2_8
+
 ```
 
 You can check for other java versions using the `module avail java` command, and load the corresponding version by including `module load java/your-version` inside your `.bashrc`.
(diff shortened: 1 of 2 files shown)


chriswyatt1:
I also think that maybe we should force singularity then to be used in the ucl_myriad config as done here:

https://github.com/nf-core/configs/blob/master/conf/eddie.config

As there is now other container engine recommended to work on myriad

docs/ucl_myriad.md
```diff
@@ -43,3 +49,13 @@ Then make sure that your bin PATH is executable, by placing the following line i
 ```bash
 export PATH=$PATH:/home/<YOUR_ID>/bin
 ```
+
+## Running an nf-core pipeline
+
+To run the pipeline, make sure to add `-profile ucl_myriad,singularity`. This will download and launch the [`ucl_myriad.config`](../conf/ucl_myriad.config) which has been pre-configured with a setup suitable for the myriad cluster, as well as use the correct container engine for myriad.
```
jfy133, line 55: Why not just set `singularity = true` in the config itself?

jfy133:
Ah what you said @chriswyatt1 :sweat_smile:

jfy133:
@nf-core-bot fix linting

docs/ucl_myriad.md
chriswyatt1: Great, we will do that :)

FernandoDuarteF:
@nf-core-bot fix linting

chriswyatt1:
@FernandoDuarteF - last commit added the CacheDir line. So should be ready now

FernandoDuarteF (approved):
Good to go

PR #785, opened 2024-10-31
Update biohpc_gen.config

nschan:
Remove resource limits params.

### DIFF
--- a/conf/biohpc_gen.config
+++ b/conf/biohpc_gen.config
@@ -20,9 +20,3 @@ charliecloud {
     enabled   = true
     writeFake = false
 }
-
-params {
-    params.max_time   = 14.d
-    params.max_cpus   = 80
-    params.max_memory = 3.TB
-}


jfy133 (approved):
That's fine but note that this will break for everything other than the most recent versions of pipelines, so it is at your own risk 😬

nschan:
Is it a pipeline or a nextflow thing? I guess in that case I will wait a couple of months...

jfy133:
Resource limits is a new native Nextflow thing to serve the same purpose as the original check_max and max_* params from nf-core.

Older versions of the nf-core pipelines will not receive updates to support this.

Note however keeping in the config will not harm current nor future pipelines. 

But otherwise, yes wait a few months of most pipelines to update to the latest template and you can remove it

Joon-Klaps:
Hi @nschan, 
Are you still testing this or do you need some help to push this over the finish line? 😄

nschan:
Hi @Joon-Klaps 

I think this can be merged, I mostly waited a bit to allow time for everyone to switch to newer nextflow versions

PR #787, opened 2024-11-05
Update dataset links and tags of nextclade to be compatible with v3+ 

Joon-Klaps:
---
name: New Config
about: A new cluster config
---

Linked to https://github.com/nf-core/viralrecon/pull/446 and https://github.com/nf-core/test-datasets/pull/1378

Please follow these steps before submitting your PR:

- [ ] If your PR is a work in progress, include `[WIP]` in its title
- [ ] Your PR targets the `master` branch
- [ ] You've included links to relevant issues, if any

Steps for adding a new config profile:

- [ ] Add your custom config file to the `conf/` directory
- [ ] Add your documentation file to the `docs/` directory
- [ ] Add your custom profile to the `nfcore_custom.config` file in the top-level directory
- [ ] Add your custom profile to the `README.md` file in the top-level directory
- [ ] Add your profile name to the `profile:` scope in `.github/workflows/main.yml`
- [ ] OPTIONAL: Add your custom profile path and GitHub user name to `.github/CODEOWNERS` (`**/<custom-profile>** @<github-username>`)

PR #789, opened 2024-11-06
Fix viralrecon main branch test using nextclade v2

Joon-Klaps:
---
name: New Config
about: A new cluster config
---

With #787, nextclade datasets were added that are compatible with v3 but not with v2. While `dev` branch might soon have v3, the `main` branch still has v2, and tests failed.  

New tags were added with the suffix `'_v3pl'` for the datasets that are compatible with nextclade's version 3 and above. 


jfy133 (approved):
Seems familiar, but OK looks simple enough

PR #801, opened 2024-11-18
VSC_UGENT: fix issue for tier1 clusters

nvnieuwk:
This fixes a small issue when running the config on tier1 clusters where `dodrio` was used as cluster name instead of the specific tier1 partition


jfy133 (approved):

nvnieuwk:
Thanks @jfy133 !

PR #803, opened 2024-11-19
Reflect changes on what jobs may run on shared partition

pontus:
Adapt to changed reality, we must now regard `MaxMemPerCPU` (which regards actual CPUs, not hyperthread units).


mashehu (approved):

aersoares81 (commented):
This should fix the problem some users are having when using the kth profile. 👍

PR #805, opened 2024-11-22
Add googlebatch config docs and update default_private_ip to false

hnawar:
---
name: googlebatch
about: Update to existing googlebatch config
---

### DIFF
--- a/conf/googlebatch.config
+++ b/conf/googlebatch.config
@@ -15,7 +15,7 @@ params {
 
 
     //networking
-    use_private_ip             = true
+    use_private_ip             = false
     // Custom VPC should be in this format 'global/networks/[custom_VPC]'
     custom_vpc                 = null
     //Custom subnet should be in this format 'regions/[GCP_Region]/subnetworks/[custom_subnet]'
--- a/docs/googlebatch.md
+++ b/docs/googlebatch.md
@@ -1,0 +1,49 @@
+# nf-core/configs: googlebatch Configuration
+
+To be used with the `googlebatch` profile by specifying the `-profile googlebatch` when running nf-core pipelines.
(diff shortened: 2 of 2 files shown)


jfy133:
@nf-core-bot fix linting

jfy133 (commented):
Hi @hnawar ,

I started an internal discussion about this when you first opened but then had sick kits along the way.

So on onehand, being a newcomer to this config, I personally don't particularly like all these extra parameters that are being added (e.g. `--project_id`, `--location`, `--use_spot` etc). These aren't supported by nf-core pipelines so every user will get a warning of an 'unknown parameter' at the beginning of every run with no validation or check it's working correctly. So even though you say they are 'required' parameters, no pipeline will actually check these.

Given it seems to be that much of this config requires a lot of user configuration, I would rather a user make their own custom config for each project/run if this is the case. 

The other issue is because Nextflow config resolution particularly is problematic when it comes to the `params` scope, meaning whatever is set in this config could be entirely overwritten if someone provides an additional custom config - so we like to avoid as much 'flexible' scopes of the pipeline as possible. 

However as a compromise as this config has already been around for a while, could you please update the config to tell validation to ignore those parameters so users at least don't get warnings of confusing 'unknown' parameters, e.g.

image

but with all parameters.

Pinging @maxulysse, @edmundmiller and @nvnieuwk who I discussed this with

jfy133:
Hi @hnawar !

We are doing a 'spring cleaning' of common nf-core github repositories.

As there has not been any activity on this, we will mark this PR to be closed by the end of the week (unless you post a comment saying you're still working on this).

 But if you see this post after it gets closed, please feel free to reopen :)

jfy133:
We will merge this for now to ensure we have docs as per discussion here: https://github.com/nf-core/configs/issues/947, but should revisit

jfy133 (approved):

PR #807, opened 2024-11-25
Extend vsc_kul_uhasselt to support gpu partitions

Joon-Klaps:
---
name: Update vsc_kul_uhasselt
about: Extend vsc_kul_uhasselt to support gpu partitions 
---

Updating the `vsc_kul_uhasselt.config` with two new profiles: 
 -  `genius_gpu`
 -  `wice_gpu`

Default `genius` and `wice` profiles, also support modules with the label `gpu` in them.

### DIFF
--- a/conf/vsc_kul_uhasselt.config
+++ b/conf/vsc_kul_uhasselt.config
@@ -55,6 +55,7 @@ profiles {
         process {
             // 768 - 65 so 65GB for overhead, max is 720000MB
             resourceLimits = [ memory: 703.GB, cpus: 36, time: 168.h ]
+            beforeScript = 'module load cluster/genius'
             clusterOptions = { "--clusters=genius --account=$tier1_project" }
 
             queue = {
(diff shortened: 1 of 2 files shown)


conf/vsc_kul_uhasselt.config
```diff
@@ -56,112 +44,137 @@ env {
...
-                        return 'batch_long'
-                    default:
-                        return 'batch'
-                    }
+                task.memory >= 175.GB ?
+                    (task.time >= 72.h ? 'dedicated_big_bigmem,dedicated_big_batch,bigmem_long' : 'bigmem') :
+                    (task.time >= 72.h ? 'batch_long' : 'batch')
+            }
+
+            withLabel: '.*gpu.*'{
```
nvnieuwk, line 67: Maybe restrict this to `gpu` only, I can see some edge cases come up if you do it like this

Joon-Klaps: mmmh... but what if non nf-core pipelines use other labels like 'process_gpu' or 'gpu_job'? @tatianawoller any  thoughts?

nvnieuwk: It's fine for me but I can see this breaking other things where the word `gpu` is also in the label

Joon-Klaps: I see what you mean but which label would have the string 'gpu' in them without referring to a GPU? Or am I overlooking something?

nvnieuwk: It probably won't happen but it can also fail when the letters `gpu` are inside another word or name, but I'll approve this, you can still fix this later if you were to encounter this issue

nvnieuwk (approved):

conf/vsc_kul_uhasselt.config
Joon-Klaps: Yeah, ok. If issues occur we'll patch

Joon-Klaps:
Thx for review 👍

PR #817, opened 2024-12-05
Utd genomics queue variable

edmundmiller:
Realized one of the memory values was off so I set them all to a variable to avoid any missed values.


jfy133 (approved):

alyssa-ab (approved):
lgtm! 🚀

PR #822, opened 2025-01-06
Update vsc_kul_uhasselt.conf -- Dynamically adjust account based on queue type

ljwharbers:
You recently added the dedicated nodes as possible partitions. However, since there is only `$SLURM_ACCOUNT` is used as the `--account`. Nextflow submissions will fail because the dedicated nodes have different `--account` requirements. 

This PR has changes to dynamically evaluate whether the outcome of the queue variable is one of the dedicated nodes and will give change the `--account` cluster option based on this. Another option is to have users specify multiple environment variables and call these instead of the hardcoded `--account` in clusterOptions like I have done now. 

Since there are issues with evaluating `queue`, I repeated the queue definition statement. It's a bit ugly so if you have other suggestions, feel free.

I have only tested this config for the dedicated vs non-dedicated CPU nodes, not for the gpu or the gpu_h100.


Joon-Klaps (approved):
Awesome, well spotted!

PR #824, opened 2025-01-07
Updated kaust profile

husensofteng:
---
name: Updated kaust profile
about: Update the main kaust config and added pipeline specific profiles for rnaseq and mag to overwrite some default limits set by those workflows. 
---

### DIFF
--- a/conf/kaust.config
+++ b/conf/kaust.config
@@ -28,8 +28,8 @@ process {
     beforeScript = 'module load singularity'
     // Max allowed resources per process on Ibex
     resourceLimits = [
-        memory: 2998.GB,
-        cpus: 128,
+        memory: 16.TB,
+        cpus: 1300,
         time: 14.d
     ]
 }
@@ -44,7 +44,7 @@ process {
         time   = { 6.h  * task.attempt }
     }
     withLabel:process_medium {
-        cpus   = { 25 * task.attempt }
+        cpus   = { 12 * task.attempt }
         memory = { 97.GB * task.attempt }
         time   = { 12.h  * task.attempt }
     }
--- a/conf/kaust.config
+++ b/conf/kaust.config
@@ -4,7 +4,6 @@ params {
     config_profile_description = 'Profile for use on King Abdullah Univesity of Science and Technology (KAUST) Ibex Cluster.'
     config_profile_contact = 'Husen Umer (@kaust.edu.sa)'
     config_profile_url = 'https://docs.hpc.kaust.edu.sa/quickstart/ibex.html'
-    save_reference = false
     igenomes_ignore = true
 }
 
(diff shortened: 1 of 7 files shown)


conf/kaust.config
```diff
@@ -4,12 +4,11 @@ params {
     config_profile_description = 'Profile for use on King Abdullah Univesity of Science and Technology (KAUST) Ibex Cluster.'
     config_profile_contact = 'Husen Umer (@kaust.edu.sa)'
     config_profile_url = 'https://docs.hpc.kaust.edu.sa/quickstart/ibex.html'
-    save_reference = false
     igenomes_ignore = true
 }
 
 // Load genome resources and assets hosted by the Bioinformatics team on IBEX cluster
-// includeConfig '/biocorelab/BIX/resources/configs/genomes.yaml'
+includeConfig '/biocorelab/BIX/resources/configs/genomes.config'
```
jfy133, line 11: This will not work because it cannot be linted/tested here as the GitHub actions nodes will not have access this file!

Please ask on #configs on slack for suggestions how to get around this

docs/kaust.md
```diff
@@ -24,20 +24,35 @@ module load nextflow
...
+Using the KAUST profile, `Docker` image(s) containing required software(s) will be downloaded, and converted to `Singularity` image(s) if needed before execution of the pipeline. To avoid downloading same images by multiple users, we provide a singularity `libraryDir` that is configured to use images already downloaded in our central container library. Images missing from our library will be downloaded to the user's directory as defined by `cacheDir`.
+
+### Accessing reference genomes on Ibex
+
+We provide a collection of reference genomes, enabling users to run workflows seamlessly without needing to download the files. To enable access to this resource, simply add the `includeConfig` line in the script below to a `nextflow.config` file under the launch directory.
```
jfy133, line 31: Are you a system administration?

It might be easier to add to all users home directory a `.nextflow/config` file which contains this (rather than users having to make their own each time), otherwise we can discuss other options.

The `.nbextflow./config` file gets implicityly loaded by all nextflow runs, which is why it's a 'nice' option

jfy133 (commented):
See my comment why tests are failing

You need some more documentation about the pipeline specific configs: https://github.com/nf-core/configs/tree/master/docs/pipeline

And also on the main README of the repo where we list

husensofteng:
> See my comment why tests are failing
> 
> You need some more documentation about the pipeline specific configs: https://github.com/nf-core/configs/tree/master/docs/pipeline
> 
> And also on the main README of the repo where we list

Thanks for the pointer, I have now added the docs.

conf/kaust.config
husensofteng: I know that is not working and I tried to find some solution through the [configs channel](https://nfcore.slack.com/archives/CGRTMASKY/p1733662786166109) but have not gone much so far. I have reposted a question, will see how it goes. 
The base line is that although it is not working with the profile but when the institute config file implicitly specified with `-c` flag then it works therefore I thought of keeping it here even though the lint test fails.

docs/kaust.md
husensofteng: Nope and I don't think they would accept that since the genomes.config file will be changed by us frequently!

husensofteng:
@jfy133 thanks for suggestion, now the config is also working after moving the file to a public github repo :)

jfy133 (approved):

jfy133:
You can merge @husensofteng !

PR #833, opened 2025-01-24
add Multiplesequencealign CRG specific config

luisas:
---
name: New Multiplesequencealign Config for CRG 
about: A new cluster config
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,8 @@ Currently documentation is available for the following pipelines within specific
   - [EVA](docs/pipeline/mag/eva.md)
 - methylseq
   - [FUB Curta](docs/pipeline/methylseq/fub_curta.md)
+- multiplesequencealign
+  - [CRG](docs/pipeline/multiplesequencealign/crg.md)
 - rnafusion
   - [HASTA](docs/pipeline/rnafusion/hasta.md)
   - [MUNIN](docs/pipeline/rnafusion/munin.md)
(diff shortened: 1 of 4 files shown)


mirpedrol (approved):
I think if we call the profile `crg` it will colide with [the general CRG config ](https://github.com/nf-core/configs/blob/5b2c6e5e229b7cf21fbd64acdc2e5627df73e845/nfcore_custom.config#L97-L99). We should find a specific name for this one, maybe `crg_msa`? 
Or we copy the same specifications to this profile, but this will be more difficult to maintain.

Otherwise LGTM :)

luisas:
> I think if we call the profile `crg` it will colide with [the general CRG config ](https://github.com/nf-core/configs/blob/5b2c6e5e229b7cf21fbd64acdc2e5627df73e845/nfcore_custom.config#L97-L99). We should find a specific name for this one, maybe `crg_msa`? Or we copy the same specifications to this profile, but this will be more difficult to maintain.
> 
> Otherwise LGTM :)

i think in other cases they also do not change the name, like here 

https://github.com/nf-core/configs/blob/master/pipeline/proteinfold.config

do u think I can leave it? I like it better so that i do not have to remember this extra config name too

mirpedrol:
Oh ok, then it doesn't override it. I see @JoseEspinosa did the proteinfold one, so if he says it's ok like this I am happy with it 🙂

PR #852, opened 2025-03-04
Add eva slurm grace cluster

maxibor:
---
name: New Config
about: A new cluster config
---

### DIFF
--- a/conf/eva.config
+++ b/conf/eva.config
@@ -3,6 +3,10 @@ params {
     config_profile_description = 'Generic MPI-EVA cluster(s) profile provided by nf-core/configs.'
     config_profile_contact     = 'James Fellows Yates (@jfy133)'
     config_profile_url         = 'https://eva.mpg.de'
+    igenomes_base              = "/mnt/archgen/public_data/igenomes"
+    max_memory                 = 2.TB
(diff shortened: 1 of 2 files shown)


jfy133 (changes requested):
I think something has gone wrong here - you've got root params twice now rather than per cluster... and this would not necessarily work because evo genetics have their own 'half' of their old cluster and wouldn't necessarily be able to access some files.

Also we are now not promoting the use of profiles within a given config because of inheritance uncertainties when specifyuing the profile order, but rather having separate configs per cluster (see `mpcdf_viper` vs `mpcdf`) for example.

So I would recommend leave `eva` as is, and making a new config called `eva_grace`

Joon-Klaps:
Hi @maxibor !

We are doing a 'spring cleaning' of common nf-core github repositories.

As there has not been any activity on this, we will mark this PR to be closed by the end of the week (unless you post a comment saying you're still working on this).

But if you see this post after it gets closed, please feel free to reopen :)

maxibor:
Hi @Joon-Klaps ,
After internal discussions with @jfy133 , we've decided to leave the eva config as is for now. I'll close the PR.

PR #854, opened 2025-03-05
Restructure pipeline configs to improve discoverability

### LINKED ISSUE #806: Restructure repo
Might break some stuff.

@jfy133 anythoughts on moving to a structure similar to the modules repo?

```console
$ tree conf/utd_ganymede/
conf/utd_ganymede/
├── utd_ganymede.config
└── utd_ganymede.md
```

Maybe a better example

```console
$ tree conf/uppmax/
conf/uppmax/
├── README.md
├── pipeline
│   ├── ampliseq.config
│   └── ampliseq.md
└── uppmax.config
```
Could nest the ampliseq one more time too.

edmundmiller:
I think restructuring the configs across the board like this would make it easier for new maintainers to find configs for pipeline for their institutions.

closes #806


lpantano:
I support this! At least for me is easy to see now.

jfy133:
This might be a good Hackathon task thinking about it (just moving stuff around), but we would need to coordinate with @mashehu as will need to update how the website parses this repo

edmundmiller:
Hey @mahesh-panchal, sorry for missing this. 

I think it's also around the pipeline-specific configs and the discoverability of that. 

I think a lot of people look for the examples, and then they find them hidden under layers and layers of directories and spread all over the place.

The way this ended up structured is that we're organizing based on the "types" of things. We're matching "this is a doc," "this is a pipeline config," and "this is an HPC system config." 

Versus "Here's all of this specific HPC system's docs and configs in one directory".

jfy133:
I don't think it's just the discoverability that would benefit from the new structure that @edmundmiller but for creation and maintenance, which the website would not help.

Alongside the examples that @edmundmiller mentioned: everything is self contained in a single directory - it's easier to cross reference your 'own' configs related to the same infra with fewer clicks and to modify in the same way (e.g. imagine doing a global search/replace in an IDE across all configs - the restriction pattern will be a single directory Vs a more complex pattern with the current system).

It also would make initialisation of templates for configs etc much simpler in the future - something that the config builder that @mirpedrol has taken over would potentially benefit from.

So /approve (if that would work here... Maybe this could be an RFC @edmundmiller )

Also I guess you could probably write a little script to update the structure across all the configs @edmundmiller ?

jfy133:
Edit: oh you did that already 🤣

ewels (approved):
Agree with the above, I like this 👍🏻 The repo is old and has grown in scope over time, this structure makes more sense now, given the content that it now creates.

I don't think that it should affect pipelines, as they only import the one file, which imports everything else (yay past-us for doing that). So it's just the website that needs updating to match. Plus this PR should include all files to rename.

Not sure that it's worthy of a hackathon project. It should be a 5 minute job with Claude Code. Not sure how complex the website update will be, but again I'd be surprised if it's that bad.

edmundmiller:
> @edmundmiller how are we with this one?

Can we piece-meal migrate stuff, or do you want everything in one go?

PR #863, opened 2025-03-12
new institutional profile for Human Technopole

edg1983:
---
name: humantechnopole
about: new institutional profile for Human Technopole
---

Hi,

I've created a new institutional profile for Human Technopole. 

Thanks!


docs/humantechnopole.md
```diff
@@ -0,0 +1,68 @@
+# nf-core/configs: Human Technopole configuration
+
+To use, run the pipeline with `-profile humantechnopole`. This will download and launch the [`humantechnopole.config`](../conf/humantechnopole.config) which has been pre-configured with a setup suitable for the Human Technopole SLURM cluster.
+
+This configuration is tested for Nextflow 23.10 and above. 
+
+## Before you use this profile
+
+1. This profile automatically load Singularity module before the script using `beforeScript` directive. Hence, when you set for example `-profile singularity`, you should be able to use singulatiry containers without any additional configuration.
```
jfy133, line 9:
```suggestion
1. This profile automatically load Singularity module before the script using `beforeScript` directive. Hence, when you set for example `-profile singularity`, you should be able to use singularity containers without any additional configuration.
```

However a user never needs to specify the singularity profile as it is already specified in your new config so they only need to specify your config and thats it. So point 2 is not necessary.

jfy133 (approved):

docs/humantechnopole.md
edg1983: Hi.

In my current configuration I only set `autoMounts = true` and some container options that are needed on our system.

In my understanding, to actually activate the singularity support, the user needs to also specify the singularity profile to set `singularity.enabled = true`.

Is this correct or am I missing something / doing something wrong here?

Thanks.

jfy133: Ah I see! Ok that makes a little more sense. Then I guess you have two options:

1. In the singularity scope just set `enabled = true` by default - is there a reason why you don't do this?
2. Change Your example to `-profile human_technopole, singularity` (without specifying your config, the whole thing is useless)

docs/humantechnopole.md
```diff
@@ -0,0 +1,68 @@
...
+#SBATCH -J job_name 
+#SBATCH -o /path/to/a/log/%x_%A.log
+#SBATCH -c 1
+#SBATCH --mem 8G
+#SBATCH -p cpuq
+
+module load nextflow/24.10.4 #Change this to other version if needed
+
+export NXF_ANSI_LOG=false
+export NXF_OPTS="-Xms8G -Xmx8G -Dnxf.pool.maxThreads=2000"
```
pontus, line 52: You likely don't need this much memory for the monitoring process, but I'd expect requesting (with `-Xms`) the same memory for the heap as the job memory request to OOM fairly often (and if it does, it's probably not a great example script).

docs/humantechnopole.md
```diff
@@ -0,0 +1,68 @@
+# nf-core/configs: Human Technopole configuration
+
+To use, run the pipeline with `-profile humantechnopole`. This will download and launch the [`humantechnopole.config`](../conf/humantechnopole.config) which has been pre-configured with a setup suitable for the Human Technopole SLURM cluster.
+
+This configuration is tested for Nextflow 23.10 and above. 
+
+## Before you use this profile
+
+1. This profile automatically load Singularity module before the script using `beforeScript` directive. Hence, when you set for example `-profile singularity`, you should be able to use singularity containers without any additional configuration.
```
pontus, line 9: I haven't checked if anything has changed, but historically, `beforeScript` has helped for using singularity in the nextflow processes, but you have still been required to have it available in the context of the monitoring process for pulling the images before.

docs/humantechnopole.md
edg1983: Yes, absolutely. Thanks for pointing this out. 
I've now adjusted this to`-Xms2G` and `-Xmx7G`. Does this look good now?

docs/humantechnopole.md
edg1983: Good point. 
I've added the corresponding module load in the main script example for the master job.

pontus (approved):

edg1983:
Should I go ahead resolving the conflicts and triggering the merge or is your team responsible for that?

pontus:
> Should I go ahead resolving the conflicts and triggering the merge or is your team responsible for that?

I think you have to resolve the conflict at least given that we don't have access in your repository.

pontus:
@nf-core-bot fix linting please

nfcore_custom.config
```diff
@@ -181,6 +181,8 @@ profiles {
     hypatia {
         includeConfig "${params.custom_config_base}/conf/hypatia.config"
     }
+    humantechnopole {
+        includeConfig "${params.custom_config_base}/conf/humantechnopole.config"
```
pontus, line 185:
```suggestion
        includeConfig "${params.custom_config_base}/conf/humantechnopole.config"
        }
```

docs/humantechnopole.md
jfy133: Yes, if you put conda after the institutional one then it should turn off singularity (if the order is correct , I think your example is the right one) :)

PR #873, opened 2025-03-26
Profiles for clusters of Alliance Canada

MareikeJaniak:
---
name: New Configs for Alliance Canada
about: Several new cluster configs for Alliance Canada HPCs
---


pontus:
Looks mostly good, but it seems a bit cumbersome having four very similar profiles. If there is something you can look at on the node to determine which cluster is being used, that's typically more user friendly I think (with the added benefit of less risk of drifting, fixes being implemented in one config and not another and so on).

It's not optimal since planned changes will make it obsolete at some point in the future, but you can possibly take some inspiration from how the UPPMAX config deals with multiple clusters https://github.com/nf-core/configs/blob/master/conf/uppmax.config.

(Roughly my understanding of planned changes referenced above is that any "execution" needs to happen in a closure - that shouldn't be so bad, but could potentially be a consideration for expensive operations.)

README.md
```diff
@@ -90,6 +90,7 @@ Currently documentation is available for the following systems:
 - [ABIMS](docs/abims.md)
 - [ADCRA](docs/adcra.md)
 - [ALICE](docs/alice.md)
+- [ALLIANCE_CANADA](docs/alliance_canada.md)
```
jfy133: Will need the 5 entries here

conf/alliance_canada_beluga.config
```diff
@@ -0,0 +1,45 @@
+params {
+    config_profile_description  = 'Alliance Canada Beluga HPC config'
+    config_profile_contact      = 'https://docs.alliancecan.ca/wiki/Technical_support'
```
jfy133, line 3: This will not render on the website, has to be a name and github handle. But it might not matter. The profile_url will get rendered howver, so if teh tech support is more important yo ucan move it there (and document the wiki page for Nextflow in the readme)

conf/alliance_canada_beluga.config
```diff
@@ -0,0 +1,45 @@
...
+}
+
+apptainer {
+  autoMounts = true
+}
+
+// Group name for resource allocation must be supplied
+params {
+    alliance_group_name = null
+}
```
jfy133, line 19: This is not recommended as you will get lots of ugly warnings in nf-core pipelines saying that unrecognised parameter.

It would be better to instead pass that to your clusterOptions via a shell environment variable.

See the aarc profile for an example

jfy133 (commented):
My only requested change is the custom parameter.

Suggestion: have you considered adding `cleanup = true` to all configs? [^1]

This means that on successful completion of the run, all intermediate files in`work/` are deleted automatically (a failed run they will be kept)

[^1]: https://nextflow.io/docs/latest/reference/config.html#unscoped-options

jfy133:
@nf-core-bot bit fix linting

jfy133:
@nf-core-bot fix linting

MareikeJaniak:
@pontus Thanks for the pointer toward the UPPMAX config - we actually have access to the cluster name via the environment variable `$HOSTNAME`. So if I'm understanding correctly, I could use this within an if statement to automatically choose the correct resource limits, right?

conf/alliance_canada_beluga.config
MareikeJaniak: Yes, I noticed that during my tests. Thanks for the pointer to the aarc profile, will check it out!

conf/alliance_canada_beluga.config
MareikeJaniak: Ok, will chat with my colleague and see what they prefer.

MareikeJaniak:
@pontus @jfy133 
I've made some changes in response to your suggestions, could you take a look when you have a chance? Thanks!

docs/alliance_canada.md
```diff
@@ -0,0 +1,11 @@
+# nf-core/configs: Alliance Canada Configuration
+
+Configuration for running nextflow on the clusters of the Digital Research Alliance of Canada. Invoke by specifying `-profile alliance_canada`.
+
+You also need to supply the name of the group under which you are working or whose resource allocation you want to use by running `export SLURM_ACCOUNT=<def-user>` before you run any nf-core pipeline. If you run nf-core frequently and always use the same resource allocation, you may find it more convenient to add the `SLURM_ACCOUNT` environment variable to your `~/.bashrc` file.
+
```
pontus, line 6: Are you using `SLURM_ACCOUNT` for other things as well? slurm by default supports `SBATCH_ACCOUNT` to pass the account (I thought it supported `SLURM_ACCOUNT` but was misremembering).

pontus (approved):
Looks good to me.

docs/alliance_canada.md
MareikeJaniak: In the documentation for the alliance canada clusters, users are encouraged to set three environment variables:
```
export SLURM_ACCOUNT=def-someuser
export SBATCH_ACCOUNT=$SLURM_ACCOUNT
export SALLOC_ACCOUNT=$SLURM_ACCOUNT
```
I picked `SLURM_ACCOUNT` here mostly because it was the first line. I'm not sure how many users actually set these, though. Users sometimes also have multiple allocations from different PIs, so specifically setting it when running a pipeline for a project might make it easier to remember to choose the right allocation.

jfy133 (approved):
You can merge when yo're ready @MareikeJaniak !

MareikeJaniak:
Thanks @pontus and @jfy133! I'm just waiting to hear back who the contact person for the profile should be.

PR #883, opened 2025-04-10
Remove pe smp when running 1 cpu

chriswyatt1:
Requested that we not use -pe smp when requesting 1 cpu, and use the -R y flag. 

Broke up the line into def lines to make it a little more readable.


jfy133 (approved):

chriswyatt1:
@nf-core-bot fix linting

PR #889, opened 2025-04-15
conf/marjorie.config: add default Slurm queue

cbrueffer:
`conf/marjorie.config`: add explicit Slurm queue since "interactive" seems to be the default now

@ashotmarg

### DIFF
--- a/conf/marjorie.config
+++ b/conf/marjorie.config
@@ -19,6 +19,7 @@ process {
         time: 240.h
     ]
     executor     = 'slurm'
+    queue        = 'compute'
     scratch      = true
     maxRetries   = 3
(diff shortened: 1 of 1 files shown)


jfy133 (approved):

pontus (approved):
Ok I guess assuming it makes sense to have the partition as a user selectable - normally it does not. 

See also `process.queue` as not using `clusterOptions` might make tuning easier if a user needs to do something.

cbrueffer:
Thanks Pontus, that seems like the better option.

ashotmarg:
Thanks for changing this, I also saw this today and had to change my config after running things on Marjorie after the update.

PR #892, opened 2025-04-21
roslin.config: Few fixes

sguizard:
---
name: roslin.config
about: Few fixes for smooth runs
---

- Remove the igenome configuration. IGMM igenome folder is not accessible for Roslin users. 
- Load singularity 3 instead of 4. Version 4 throw squashfs / FUSE error on new nodes. 
- Add logic to define the singularity cache directory. Users who don't use compute project can't access the shared singularity cache.

### DIFF
--- a/conf/roslin.config
+++ b/conf/roslin.config
@@ -16,9 +16,9 @@ process {
 
     // This will override all jobs clusterOptions
     // This is necessary to allow jobs to run on Eddie for many users
-    // For each job, we add an extra 4 Gb of memory.
-    // For example, the process asked 16 Gb of RAM (task.memory). The job will reserve 20 Gb of RAM.
-    // The process will still use 16 Gb (task.memory) leaving 4 Gb for other system processes.
+    // For each job, we add an extra 8 Gb of memory.
+    // For example, the process asked 16 Gb of RAM (task.memory). The job will reserve 24 Gb of RAM.
+    // The process will still use 16 Gb (task.memory) leaving 8 Gb for other system processes.
     // This is very useful any JAVA programs which allocate task.memory RAM for its Virtual Machine
     // Also it leaves enough memory for singularity to unpack images.
     clusterOptions = {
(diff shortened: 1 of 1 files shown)


conf/roslin.config
```diff
@@ -56,6 +52,10 @@ singularity {
     runOptions   = '-p -B "$TMPDIR"'
     enabled      = true
     autoMounts   = true
-    cacheDir     = '/exports/cmvm/eddie/eb/groups/alaw3_eb_singularity_cache'
+    // Define the singularity cache directory depending on the presence of the NFX_SGE_PROJECT variable
+    // User without compute project can't access to the shared cache directory.
+    // So, they need to store singularity images into the work directory.
+    cacheDir     = System.getenv('NFX_SGE_PROJECT') ?
```
pontus, line 58: `libraryDir` instead? (In which case one might not need the conditional?)

conf/roslin.config
```diff
@@ -34,19 +34,15 @@ process {
     maxErrors = '-1'
     maxRetries = 3
 
+    // Force singularity version 3 as version 4 cause FUSE/squashfs issues on Eddie
     beforeScript =
     """
     . /etc/profile.d/modules.sh
-    module load singularity
+    module load igmm/apps/singularity/3
```
pontus, line 41: I'd strongly recommend trying to resolve this, the last 3 release is almost two years old and there's been a steady stream of fixes since.

pontus (approved):

conf/roslin.config
sguizard: The cacheDir better fit to our setup. It convenient for us to allow user to add new images when they run a new pipeline / new version. 
Thanks for the suggestion.

conf/roslin.config
sguizard: We have updated our nodes very recently. 
The IS are working on this issue. 
While waiting for a resolution, non nextflow power user struggle to debug this singularity. 
I'll switch back as soon as singularity 4 is fixed ;)

sguizard:
Many thanks for the review and the suggestions!

PR #910, opened 2025-06-10
init fred_hutch profile

derrik-gratz:
---
name: New Config
about: A new cluster config
---

  - Is this still needed? I don't see this mentioned on the README or see other profiles on there
- [x] OPTIONAL: Add your custom profile path and GitHub user name to `.github/CODEOWNERS` (`**/<custom-profile>** @<github-username>`)


conf/fred_hutch.config
```diff
@@ -0,0 +1,83 @@
...
+    // Maximum jobs per second
+    submitRateLimit = "10/1sec"
+}
+
+apptainer {
+    enabled = true
+    // If the user does not provide a scratch directory, use their home dir
+    cacheDir = params.scratch ? "${params.scratch}/.apptainer/cache" : "${System.getenv('HOME')}/.apptainer/cache"
+    // Checks if TMPDIR is defined in the environment before mounting to avoid breaks
+    runOptions = "--containall " + (System.getenv('TMPDIR') ? "-B \$TMPDIR" : "")
```
pontus, line 72: Normally I think it's a good idea to have local filesystems that aren't home bound in, it's very rare something mistakenly picks up libraries from there and being able to have external (shared) references is worth a lot with many pipelines.

.github/CODEOWNERS
```diff
@@ -5,3 +5,4 @@
 **/utd_ganymede** @edmundmiller @alyssa-ab
 **/unsw_katana** @jscgh
 **/seadragon** @jiawku
+**/fred_hutch** @derrik-gratz
```
pontus, line 8: Probably need to join the github org for this not to complain, see https://slack.com/shortcuts/Ft07NJUA4R0B/1549d7fef819a7d172c198f24ba3db11 (GitHub invitation request in #github-invitations in the nf-core slack if the link doesn't work).

conf/fred_hutch.config
```diff
@@ -0,0 +1,83 @@
...
+    // How many jobs can be submitted at once
+    queueSize = 16
+    // Maximum jobs per second
+    submitRateLimit = "10/1sec"
+}
+
+apptainer {
+    enabled = true
+    // If the user does not provide a scratch directory, use their home dir
+    cacheDir = params.scratch ? "${params.scratch}/.apptainer/cache" : "${System.getenv('HOME')}/.apptainer/cache"
```
pontus: This seems off, a scratch directory would be something you probably don't reuse between runs whereas the (nextflow) apptainer cache directory would be something you'd want to reuse.

By default it will also be in the work directory, so if you don't want to set it, that might make more sense?

conf/fred_hutch.config
```diff
@@ -0,0 +1,83 @@
...
+    config_profile_name        = 'Fred Hutch'
+    config_profile_description = 'Fred Hutch Cancer Center HPC profile'
+    config_profile_contact     = 'dgratz@fredhutch.org'
+    config_profile_url         = 'https://sciwiki.fredhutch.org/'
+    // Same as resourceLimits, but for older pipelines
+    max_memory = 700.GB
+    max_cpus = 36
+    max_time = 720.h
+    // Assumes user will provide a scratch dir at runtime
+    scratch = null
```
pontus: Naming, this might easily be confused with the nextflow `scratch` functionality which is another thing.

conf/fred_hutch.config
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// If the user does not provide a scratch directory, use their home dir
+workDir = params.scratch ? "${params.scratch}" : "${System.getenv('HOME')}/nextflow"
```
pontus: This is not exactly wrong but probably confusing for someone used to nextflow, is there a problem with the nextflow default behaviour?

derrik-gratz: we were thinking it would cause pileup of intermediate files in disparate locations which would be hard to clean, while having a single location would allow easier cleaning. Happy to hear arguments against this

conf/fred_hutch.config
derrik-gratz: I agree that it should not be in a scratch dir, we will at least remove that option. The motivation for it defaulting to `home` if scratch was not provided was to have all the apptainer images in a central location that was easy to clean, as it is also the default location apptainer. I realize you are not an apptainer dev, but please correct me if I'm wrong: the apptainer cache will be specific to the user, so having a shared location for the cache will not allow users to re-use images created by other users right?

.github/CODEOWNERS
derrik-gratz: I have been added as I also plan to pr a module!

conf/fred_hutch.config
derrik-gratz: `--contain-all` was something I lifted from some older documentation and I think I had the inverse understanding of it. I'll talk to some folks and see if there was rationale for that

derrik-gratz:
@sminot any thoughts on the `--contain-all` option for Apptainer? Was that a security thing? I lifted it from the [wiki](https://sciwiki.fredhutch.org/datademos/on_gizmo/#nextflow-configuration---gizmo)

sminot:
Yes, `--containall` was added as a general recommendation to make it more similar in execution to Docker, which doesn't add all of the filesystem mountpoints that Apptainer is set up to do by default.

conf/fred_hutch.config
sminot: I think the behavior is good, but for the implementation I think it's easier to just have the default value for the `scratch` param set to `$HOME/nextflow`, rather than using the ternary operator

derrik-gratz:
> Yes, `--containall` was added as a general recommendation to make it more similar in execution to Docker, which doesn't add all of the filesystem mountpoints that Apptainer is set up to do by default.

Are you in favor of keeping it?

sminot:
> > Yes, `--containall` was added as a general recommendation to make it more similar in execution to Docker, which doesn't add all of the filesystem mountpoints that Apptainer is set up to do by default.
> 
> Are you in favor of keeping it?

Yes, I am

docs/fred_hutch.md
```diff
@@ -0,0 +1,22 @@
+# nf-core/configs: Fred Hutch Cancer Center Configuration
+
+Configuration file to run nf-core pipelines on the cluster of the [Fred Hutchinson Cancer Center](https://www.fredhutch.org/).
+
+# Before running the pipeline
+
+You may need to load the Nextflow module or otherwise have a Nextflow executable in your path. Apptainer is also required, but is specified in the default `beforeScript` of this configuration, so you shouldn't need to load the module manually, but if you encounter issues with containers, consider loading it yourself.
```
pontus: `beforeScript` doesn't help with pulling containers which (for https specified ones) happen before the job is submitted, so it might be good to imply it stronger.

conf/fred_hutch.config
pontus: For clarity; there's an apptainer cache and a nextflow cache of apptainer images. Maybe I should have mentioned that more clearly, but often, it doesn't make sense to put the nextflow cache into the first one.

As for sharing, it "works" but nextflow will expect to be able to write to the cachedirectory for pulling.

There's also `singularity.libraryDir` / `apptainer.libraryDir` which is useful for having a central image repository (where a trusted group can make sure most images are available).

conf/fred_hutch.config
pontus: Obviously your choice, my hunch is that it's somewhat confusing compared to how most use (and e.g. what `nextflow clean` does and so on) but certainly workable.

.github/CODEOWNERS
pontus: @jfy133 I don't recall, does something need to be done on the repo here to not trigger failure with derrik-gratz as codeowner?

jfy133 (approved):

jfy133:
LGTM  @derrik-gratz !

PR #921, opened 2025-06-25
Add gpu queue to M3 Cluster config

famosab:
---
name: New Config
about: A new cluster config
---

### DIFF
--- a/conf/m3c.config
+++ b/conf/m3c.config
@@ -16,9 +16,19 @@ process {
         time: 14.d
     ]
     executor         = 'slurm'
-    queue            = { task.time > 23.h ? 'cpu3-long' : (task.memory > 460.GB || task.cpus > 64 ? 'cpu2-hm' : 'cpu1') }
     scratch          = 'true'
     containerOptions = '--bind $TMPDIR'
+    queue = {
+        if (task.time > 23.h) {
+            return 'cpu3-long'
+        } else if (task.label && task.label.toString().contains('gpu')) {
+            return 'gpu-a30'
+        } else if (task.memory > 460.GB || task.cpus > 64) {
+            return 'cpu2-hm'
+        } else {
+            return 'cpu1'
+        }
+    }
 }
 
 params {


mashehu (approved):

conf/m3c.config
```diff
@@ -16,9 +16,29 @@ process {
         time: 14.d
     ]
     executor         = 'slurm'
-    queue            = { task.time > 23.h ? 'cpu3-long' : (task.memory > 460.GB || task.cpus > 64 ? 'cpu2-hm' : 'cpu1') }
     scratch          = 'true'
     containerOptions = '--bind $TMPDIR'
+    queue = {
+        if (task.label && task.label.toString().contains('gpu')) {
+            return 'gpu-a30'
+        } else if (task.time > 24.h) {
```
skrakau, line 24: Why did you change it? could this be a problem for tasks taking 24h?

famosab: Because I found that with task.time > 23 I always entered the -long node because tasks are often submutted with time 24h

skrakau: but if 24h tasks are not allowed on 'cpu2-hm' and 'cpu1' (which was the case in the past if I remember correctly), then that would be correct, or not?

skrakau: if it get's only killed when > 24h, then it makes sense changing it as you did (not sure how the exact setting was)

famosab: I understood the cluster description as that 24h tasks is fine it just gets killed above that limit. https://wiki.m3c.uni-tuebingen.de/en/Slurm/Overview

pontus: The normal behaviour is to reject jobs that do not fit rather than allowing them to start. Should be easy enough to test (submit a job with `-t 1-00:00 -p cpu1` and see if it's accepted or not.

(It's possible they tweaked the limits so you can e.g. run up to 24 hours and 5 minutes in the 24-hour queue. If not, the config here should be fixed.)

famosab: How would I correctly specifiy this so that I do not run into using cpu-long for tasks that are submitted with exactly 24h?

famosab: `srun -t 1-00:00 -p cpu1 --pty bash` for example works

PR #931, opened 2025-07-08
Fix error in recently updated cfc and m3c configs (final version)

nadnein:
---
name: Update CFC & M3 cluster config
about: Final inclusion of co2footprint parameters
---

In the previous merged PR, the `timestamp` variable was wrongly set in the output file names definition. It is now fixed.

### DIFF
--- a/conf/cfc.config
+++ b/conf/cfc.config
@@ -39,19 +39,17 @@ params {
  * `-plugins nf-co2footprint@<VERSION>` parameter in your nextflow call.
  */
 co2footprint {
+    String co2Timestamp = new java.util.Date().format( 'yyyy-MM-dd_HH-mm-ss')
+
     // Determine timestamp suffix
-    def timestampSuffix = params.get('trace_report_suffix') ?: (this.hasProperty('trace_timestamp') ? this.trace_timestamp : '')
-    if (timestampSuffix) {
-        timestamp = timestampSuffix
-        timestampSuffix = "_${timestampSuffix}"
-    }
+    String timestampSuffix = params.get('trace_report_suffix') ?: (this.hasProperty('trace_timestamp') ? this.trace_timestamp : co2Timestamp)
 
     // File parameters
     if (params.containsKey('outdir')) {
         def infoDir = "${params.outdir}/pipeline_info"
-        traceFile   = "${infoDir}/co2footprint_trace${timestampSuffix}.txt"
-        reportFile  = "${infoDir}/co2footprint_report${timestampSuffix}.html"
-        summaryFile = "${infoDir}/co2footprint_summary${timestampSuffix}.txt"
+        traceFile   = "${infoDir}/co2footprint_trace_${timestampSuffix}.txt"
+        reportFile  = "${infoDir}/co2footprint_report_${timestampSuffix}.html"
+        summaryFile = "${infoDir}/co2footprint_summary_${timestampSuffix}.txt"
     }
 
     // Standard parameters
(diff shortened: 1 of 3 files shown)


JosuaCarl (commented):
I would replace`def` with `String` where possible. I know it's redundant, but I find it more readable with types.

conf/cfc.config
```diff
@@ -39,20 +39,22 @@ params {
...
 co2footprint {
-    // Defines the timestamp model that is used as a suffix of the file
-    String co2_timestamp = params.get('trace_report_suffix') ?: this.hasProperty('trace_timestamp') ? this.getProperty('trace_timestamp') : null
-    if (co2_timestamp != null) {
-        timestamp = co2_timestamp
+    // Determine timestamp suffix
+    def timestampSuffix = params.get('trace_report_suffix') ?: (this.hasProperty('trace_timestamp') ? this.trace_timestamp : '')
+    if (timestampSuffix) {
+        timestamp = timestampSuffix
+        timestampSuffix = "_${timestampSuffix}"
```
skrakau: adding this "_" here extra, doesn't make the code more readable for me in comparison to before

conf/cfc.config
```diff
@@ -39,20 +39,22 @@ params {
  * `-plugins nf-co2footprint@<VERSION>` parameter in your nextflow call.
  */
 co2footprint {
-    // Defines the timestamp model that is used as a suffix of the file
-    String co2_timestamp = params.get('trace_report_suffix') ?: this.hasProperty('trace_timestamp') ? this.getProperty('trace_timestamp') : null
-    if (co2_timestamp != null) {
-        timestamp = co2_timestamp
+    // Determine timestamp suffix
+    def timestampSuffix = params.get('trace_report_suffix') ?: (this.hasProperty('trace_timestamp') ? this.trace_timestamp : '')
```
skrakau: So this means currently, that if there is no (known) timestamp given already, `''` will be used for the plugin output files. Shouldn't then a new timestamp be created? 
Not sure how often this occurs, but for old nf-core pipeline versions this is the case.

JosuaCarl: There will be a timestamp, from the Groovy code. But only if the file names are not overwritten. I think we should talk about how to handle the file names/paths in general. The current approach seems messy to me.

conf/cfc.config
JosuaCarl: But it avoids an extra if-case

conf/cfc.config
```diff
@@ -39,17 +39,20 @@ params {
...
 
     // File parameters
     if (params.containsKey('outdir')) {
-        traceFile   = "${params.outdir}/pipeline_info/co2footprint_trace_${co2_timestamp}.txt"
-        reportFile  = "${params.outdir}/pipeline_info/co2footprint_report_${co2_timestamp}.html"
-        summaryFile  = "${params.outdir}/pipeline_info/co2footprint_summary_${co2_timestamp}.txt"
+        String infoDir = "${params.outdir}/pipeline_info"
+        traceFile   = "${infoDir}/co2footprint_trace_${timestampSuffix}.txt"
+        reportFile  = "${infoDir}/co2footprint_report_${timestampSuffix}.html"
+        summaryFile = "${infoDir}/co2footprint_summary_${timestampSuffix}.txt
```
nadnein:
```suggestion
        summaryFile = "${infoDir}/co2footprint_summary_${timestampSuffix}.txt"
```

PR #944, opened 2025-08-18
Update Sanger config to handle strict syntax

prototaxites:
Updates `sanger.conf` to handle the new strict config syntax (https://nextflow.io/docs/latest/strict-syntax.html#configuration-syntax).

- moves farm22/tol22 specific config to separate subconfig files
- these config files are included using a closure which first determines which cluster you are on
- Add default values for cpus, memory, and time (otherwise processes fail if these are not set)
- moved `executor.poolSize` to `poolSize`, which is the correct name for this option

Have tested this on tol22 and it seems to work correctly.

### DIFF
--- a/conf/sanger.config
+++ b/conf/sanger.config
@@ -13,7 +13,7 @@ process {
     executor = 'lsf'
 
     // Processes without any explicit labels will fail to run on LSF
-    // Set low values as defaults to have a default value 
+    // Set low values as defaults to have a default value
     cpus   = 1
     memory = 6.Gb
     time   = 1.h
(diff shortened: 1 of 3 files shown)


conf/sanger.config
```diff
@@ -55,49 +56,30 @@ process {
...
-    process {
-        resourceLimits = [
-            memory: 1.4.TB,
-            cpus: 64,
-            time: 89280.min
-        ]
+    try {
+        clustername = ['/bin/bash', '-c', 'lsid | awk \'$0 ~ /^My cluster name is/ {print $5}\''].execute().text.trim()
+    } catch (java.io.IOException e) {
+        System.err.println("WARNING: Could not run lsid to determine current cluster, defaulting to farm")
```
pontus:
```suggestion
        System.err.println("WARNING: Could not run lsid to determine current cluster, defaulting to "+clustername)
```

having a warning about defaulting to what's not used as default might be slightly confusing.

pontus (approved):

muffato (approved):

PR #950, opened 2025-08-28
Amend qmul_apocrita config readme for best practices

cpearse:
Since this isn't officially supported by the QMUL ITSR team (which I am a staff member of), can we please make this much clearer? Users of the cluster are approaching us for support as if this is official.

Also included are several suggested adjustments that align setup with our best practices (removing piping `curl` to `bash`, using Private Modules, not using `~/.bashrc` etc.)

### DIFF
--- a/conf/qmul_apocrita.config
+++ b/conf/qmul_apocrita.config
@@ -1,6 +1,6 @@
 params {
 
-    config_profile_description = 'Queen Mary Universtiy of London'
+    config_profile_description = '(Unofficial) Queen Mary University of London'
     config_profile_contact     = 'Simon Murray (simon . murray AT ucl . ac . uk)'
     config_profile_url         = 'https://docs.hpc.qmul.ac.uk/'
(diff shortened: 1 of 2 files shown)


conf/qmul_apocrita.config
```diff
@@ -1,6 +1,6 @@
 params {
 
-    config_profile_description = 'Queen Mary Universtiy of London'
+    config_profile_description = 'Queen Mary University of London'
```
jfy133:
```suggestion
    config_profile_description = '(Unofficial) Queen Mary University of London'
```

jfy133 (approved):
Absolutely, I've also made another suggestion to make it even clearer (the description gets printed at the beginning of each pipeline run.

conf/qmul_apocrita.config
cpearse: Great, I've squashed that into a new commit.

PR #951, opened 2025-08-29
Restructure all the configs

edmundmiller:
- **refactor: Add example moves**
- **chore: Add docs on how this is used**
- **refactor(uppmax): reorganize directory structure**
- **fix(uppmax): update config path in nfcore_custom.config**
- **feat: add automation script for cluster config reorganization**
- **refactor: reorganize sanger cluster and improve automation script**
- **docs: document new cluster directory structure and migration process**
- **fix: update GitHub workflows for new directory structure**
- **refactor: migrate all 134 remaining clusters to new directory structure**


jfy133:
@edmundmiller I would need time to properly test this (which won't come for a while....unless we look during the retreat) but i. The meantime are the AI scripts meant to stay in here?

Also I can't see from my phone as too many files - but have you updated the auto generation action for new configs?

PR #952, opened 2025-09-01
Updated resource limits for Janelia's LSF cluster

krokicki:
Janelia's compute cluster resources have been upgraded since this config was first published. 

I have updated the config with the new max and default values for cpu/memory/runtime.


krokicki:
@pontus Thanks a lot for pointing that out! It was not intentional, I misremembered how those work. It's fixed now.

pontus (approved):

PR #960, opened 2025-09-25
new RKI config

MarieLataretu:
---
name: New Config
about: A new cluster config
---


conf/rki.config
```diff
@@ -0,0 +1,52 @@
+//Profile config names for nf-core/configs
+params {
+    config_profile_description = 'Generic RKI HPC profile provided by nf-core/configs.'
+    config_profile_contact = 'Marie Lataretu (@MarieLataretu)'
+    config_profile_url = 'https://www.rki.de/'
+    conda_cache_dir = 'conda'
```
jfy133: Is this correct? This would be local to the pipeline run... (and would not exist)?

conf/rki.config
```diff
@@ -0,0 +1,52 @@
...
+}
+
+trace {
+    fields = 'task_id,hash,workdir,native_id,name,tag,status,attempt,exit,submit,duration,realtime,%cpu,peak_rss,peak_vmem,rchar,wchar,container,module'
+}
+
+profiles {
+    // To use singularity, use nextflow run -profile rki,singularity
+    singularity {
+        singularity {
```
jfy133: No cache dir for this one?

conf/rki.config
```diff
@@ -0,0 +1,52 @@
...
+            enabled = true
+            autoMounts = true
+            envWhitelist = "HTTPS_PROXY,HTTP_PROXY,http_proxy,https_proxy,FTP_PROXY,ftp_proxy"
+        }
+    }
+    // To use conda, use nextflow run -profile rki,conda
+    conda {
```
jfy133: I think this is very dangerous, because we already have a top level `conda` profile in our template.

I think just set `conda.cacheDir` in your base config

docs/rki.md
```diff
@@ -0,0 +1,26 @@
...
+
+All nf-core pipelines have been successfully configured for use on the HPC at RKI.
+
+To use, run the pipeline with `-profile PROFILENAME`. This will download and launch the [`profile.config`](../conf/profile.config) which has been pre-configured with a setup suitable for the HPC. Using this profile, a docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.
+
+## Before running the pipeline
+
+### Singularity
+
+We highly recommend specifying a location of a cache directory to store singularity images (so you re-use them across runs, and not pull each time), by specifying the location with the `$NXF_SINGULARITY_CACHEDIR` bash environment variable in your `~/.bashrc`, or in your current session:
```
jfy133, line 22: Can you not set up a global directory?

jfy133 (commented):
half review in between nf-core retreat sessions

conf/rki.config
```diff
@@ -0,0 +1,52 @@
+//Profile config names for nf-core/configs
+params {
+    config_profile_description = 'Generic RKI HPC profile provided by nf-core/configs.'
+    config_profile_contact = 'Marie Lataretu (@MarieLataretu)'
+    config_profile_url = 'https://www.rki.de/'
+    conda_cache_dir = 'conda'
+}
+
+process {
+    executor = 'slurm'
```
pontus, line 13: Maybe some `resourceLimits` here is helpful?

conf/rki.config
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+    // To use mamba, use nextflow run -profile rki,mamba
+    mamba {
+        conda {
+            enabled = true
+            useMamba = true
+            cacheDir = params.conda_cache_dir
+        }
+    }
```
pontus: These mostly seem to mirror what's in the nf-core template?

pontus: I also think it makes sense for a profile to be opinionated and do the best thing (extremely likely singularity here).

docs/rki.md
```diff
@@ -0,0 +1,26 @@
+# nf-core/configs: PROFILE Configuration
+
+All nf-core pipelines have been successfully configured for use on the HPC at RKI.
+
+To use, run the pipeline with `-profile PROFILENAME`. This will download and launch the [`profile.config`](../conf/profile.config) which has been pre-configured with a setup suitable for the HPC. Using this profile, a docker image containing all of the required software will be downloaded, and converted to a Singularity image before execution of the pipeline.
```
pontus: Missing adaption?

docs/rki.md
MarieLataretu: Currently, we have only group-wide shared directories, but not an institute-wide one! 

We try to set up something institute-wide, but I can't say when and how this will look. For now, I'd go with this solution to make it flexible for all.

conf/rki.config
MarieLataretu: Set via `$NXF_SINGULARITY_CACHEDIR` (see readme/below)

conf/rki.config
MarieLataretu: Hm, true. 

So the thing is, quite some people use mamba/conda and want to have a custom directory for the cache.

The default value might be suboptimal, but this would allow users to set the conda cache dir from the CLI. Or is there another way, that I'm not aware of?

conf/rki.config
MarieLataretu: We have also mamba/conda users who want to be able to set the cache dir - I tried to catch all use cases with this config 🤔

docs/rki.md
MarieLataretu: Thx, fixed!

conf/rki.config
pontus: Understood, but again, for a profile I feel it's okay to be opinonated and avoid things that can cause issues (if people have other opinions, they can just not use the profile).

qiyubio: My two cents: adding mamba/conda on HPC profile only bring troubles for debugging when things going south.

qiyubio: Also pixi support is on the horizon: https://github.com/nextflow-io/nextflow/issues/6092

conf/rki.config
pontus: You don't need to set a default at all (and again, it's fine to be opinionated rather than wasting time trying to support things that are prone to breaking).

MarieLataretu: removed in favor of `NXF_CONDA_CACHEDIR`

conf/rki.config
MarieLataretu: resolved with `NXF_CONDA_CACHEDIR`

conf/rki.config
MarieLataretu: I removed all the profiles and added `NXF_CONDA_CACHEDIR` to the docs 👍

conf/rki.config
MarieLataretu: Is that not covered by the template?

pontus: I don't think I understand what template but probably no.

MarieLataretu: added `resourceLimits` 👍

PR #1015, opened 2026-02-18
FIX: Strict syntax for vsc_kul_hasselt

nvnieuwk:
This PR makes the config compatible with strict syntax.

@Joon-Klaps would you mind testing this out on the cluster? I'm not 100% sure the use of closures here is correct (even though it should work?)

You can run a quick test using the test profile of `rnavar` since that pipeline is strict syntax compatible.

`NXF_VER=26.01.1-edge NXF_SYNTAX_PARSER=v2 nextflow run nf-core/rnavar -r dev -profile test -c <path-to-this-config>`

Feel free to also test it with a pipeline you run without the strict syntax. To make sure that also still works.


jfy133 (approved):
That's complicated, but I trust you 😬

maxulysse:
@Joon-Klaps any update on this one?

Joon-Klaps:
No, we still havent found a good way to dynamically select the correct partition. I also cannot find the time atm to work on this.

maxulysse:
Replaced by https://github.com/nf-core/configs/pull/1112

PR #1093, opened 2026-04-22
embl_hd: Add libraryDir and update executor configuration

fullama:
Add libraryDir configuration for Apptainer and update executor block.

### DIFF
--- a/conf/embl_hd.config
+++ b/conf/embl_hd.config
@@ -48,11 +48,10 @@ process {
 }
 
 executor {
-  $slurm {
+    name = 'slurm'
     queueSize = 200
     submitRateLimit = "10/1sec"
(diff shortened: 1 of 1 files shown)


pontus (approved):

pontus:
Normally the PR submitter merges, but since we're working on shifting to strict syntax you may not be able to - if that's the case, let us know and we can merge.

fullama:
> Normally the PR submitter merges, but since we're working on shifting to strict syntax you may not be able to - if that's the case, let us know and we can merge.

thanks for taking a look! but yeah it wont let me due to failing tests

fullama (commented):
maybe that wasnt the best change

PR #1094, opened 2026-04-22
Fix link to sanger.config in sanger documentation

jfy133:
Updated the link to the sanger.config file so it can get properly rendered on the website, to address: https://github.com/nf-core/website/issues/3702


priyanka-surana (approved):
Thanks 👍🏼

PR #1102, opened 2026-04-29
[WIP] Update Cambridge config for current CSD3 partitions

RaqManzano:
---
name: Update Cambridge CSD3 Config
about: Updating Cambridge CSD3 cluster config
---

## Summary

This PR updates the Cambridge CSD3 institutional profile to reflect the current partitions and simplifies the configuration logic based on feedback from the nf-core team.

## Changes

- update `conf/cambridge.config` with current partition limits:
  - `icelake`
  - `icelake-himem`
  - `sapphire`
- set `icelake` as the default partition
- keep `--partition` as a user override
- infer walltime limits from the SLURM account name:
  - accounts containing `-SL3-` get a `12.h` cap
  - otherwise default to `36.h` for SL1 / SL2
- add schema validation ignores for config-specific parameters
- refresh `docs/cambridge.md`:
  - update install instructions
  - document partition selection
  - explain `screen` / `tmux` usage
  - add a note recommending `srun` / `sbatch` for large runs where the Nextflow manager can become memory-heavy
- add Cambridge to `.github/CODEOWNERS`
- add Cambridge-specific module initialization and SLURM executor tuning to improve job submission stability on CSD3

## Ackowledgements

Many thanks to @pontus, @jfy133, @tdanhorn and @maxulysse  for their initial feedback in the nf-core channel.

### DIFF
--- a/conf/cambridge.config
+++ b/conf/cambridge.config
@@ -59,11 +59,11 @@ singularity {
 process {
     resourceLimits = params.csd_selected
 
-    beforeScript = '''
+    beforeScript = """
     . /etc/profile.d/modules.sh
     module purge
-    module load rhel8/default-icl
-    '''
+    module load rhel8/default-${params.partition == 'sapphire' ? 'sar' : 'icl'}
+    """
 
     executor       = 'slurm'
     queue          = params.partition
--- a/conf/cambridge.config
+++ b/conf/cambridge.config
@@ -18,42 +18,38 @@ params {
     project   = null
 
     // Compatibility with nf-core schema validation across pipeline versions.
-    schema_ignore_params         = 'partition,project,max_memory,max_cpus,max_time,validationSchemaIgnoreParams'
-    validationSchemaIgnoreParams = 'partition,project,max_memory,max_cpus,max_time,schema_ignore_params,validationSchemaIgnoreParams'
+    schema_ignore_params         = 'partition,project,max_memory,max_cpus,max_time,csd_time,csd_parts,csd_selected,validationSchemaIgnoreParams'
+    validationSchemaIgnoreParams = 'partition,project,max_memory,max_cpus,max_time,csd_time,csd_parts,csd_selected,schema_ignore_params,validationSchemaIgnoreParams'
 }
 
 validation {
-    ignoreParams = ['partition', 'project', 'max_memory', 'max_cpus', 'max_time', 'schema_ignore_params', 'validationSchemaIgnoreParams']
+    ignoreParams = ['partition', 'project', 'max_memory', 'max_cpus', 'max_time', 'csd_time', 'csd_parts', 'csd_selected', 'schema_ignore_params', 'validationSchemaIgnoreParams']
 }
 
-params.max_memory = {
-    if (params.partition == 'icelake') {
-        return 256.GB  // this is a whole node
-    }
-    if (params.partition == 'icelake-himem' || params.partition == 'sapphire') {
-        return 512.GB // this is a whole node
-    }
-
-    System.err.println("ERROR: cambridge params.partition must be one of 'icelake', 'icelake-himem', or 'sapphire' (got '${params.partition}').")
-    System.exit(1)
+params.csd_time = {
+    params.project?.toUpperCase()?.contains('-SL3-') ? 12.h : 36.h
 }.call()
 
-params.max_cpus = {
-    if (params.partition == 'sapphire') {
-        return 112
-    }
-    if (params.partition == 'icelake' || params.partition == 'icelake-himem') {
-        return 76
+params.csd_parts = [
+    icelake        : [memory: 256.GB, cpus: 76, time: params.csd_time],
+    'icelake-himem': [memory: 512.GB, cpus: 76, time: params.csd_time],
+    sapphire       : [memory: 512.GB, cpus: 112, time: params.csd_time]
+]
+
+params.csd_selected = {
+    def selected = params.csd_parts[params.partition]
+
+    if (!selected) {
+        System.err.println("ERROR: cambridge params.partition must be one of 'icelake', 'icelake-himem', or 'sapphire' (got '${params.partition}').")
+        System.exit(1)
     }
 
-    System.err.println("ERROR: cambridge params.partition must be one of 'icelake', 'icelake-himem', or 'sapphire' (got '${params.partition}').")
-    System.exit(1)
+    selected
 }.call()
 
-params.max_time = {
-    def project_name = params.project?.toUpperCase()
-    return project_name?.contains('-SL3-') ? 12.h : 36.h
-}.call()
+params.max_memory = params.csd_selected.memory
+params.max_cpus   = params.csd_selected.cpus
+params.max_time   = params.csd_selected.time
 
 singularity {
     enabled    = true
@@ -61,11 +57,7 @@ singularity {
 }
 
 process {
-    resourceLimits = [
-        memory: params.max_memory,
-        cpus: params.max_cpus,
-        time: params.max_time
-    ]
+    resourceLimits = params.csd_selected
 
     executor       = 'slurm'
     queue          = params.partition
--- a/conf/cambridge.config
+++ b/conf/cambridge.config
@@ -59,7 +59,22 @@ singularity {
 process {
     resourceLimits = params.csd_selected
 
+    beforeScript = '''
+    . /etc/profile.d/modules.sh
+    module purge
+    module load rhel8/default-icl
+    '''
+
     executor       = 'slurm'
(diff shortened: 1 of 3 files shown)


docs/cambridge.md
```diff
@@ -7,81 +7,99 @@ and converted to a Singularity image or a Singularity image downloaded directly
...
+more conservative recommendation here.
 
 ### Set up Singularity cache
 
 Singularity allows the use of containers and will use a caching strategy. First, you might want to set the `NXF_SINGULARITY_CACHEDIR` bash environment variable, pointing at your hpc-work location. If not, it will be automatically assigned to the current directory.
 
 ```
 # do this once per login, or add these lines to .bashrc
-export NXF_SINGULARITY_CACHEDIR=/home/<username>/rds/hpc-work/path/to/cache/dir
+export NXF_SINGULARITY_CACHEDIR=/home/<username>/singularity
```
pontus: This could be using `$USER` to help people that don't understand they need to replace `<username>`? (https://nf-co.re/docs/specifications/configs/documentation#examples-in-documentation)

pontus (approved):

docs/cambridge.md
RaqManzano: I changed it now to `$HOME`

conf/cambridge.config
```diff
@@ -1,28 +1,80 @@
...
+    config_profile_description = 'Cambridge HPC CSD3 cluster profile.'
+    config_profile_contact     = 'Raquel Manzano (rm889@cam.ac.uk) and Andries van Tonder (ajv37@cam.ac.uk)'
+    config_profile_url         = 'https://docs.hpc.cam.ac.uk/hpc'
+
+    partition = 'icelake'
+    project   = null
+
+    // Compatibility with nf-core schema validation across pipeline versions.
+    schema_ignore_params         = 'partition,project,max_memory,max_cpus,max_time,csd_time,csd_parts,csd_selected,validationSchemaIgnoreParams'
+    validationSchemaIgnoreParams = 'partition,project,max_memory,max_cpus,max_time,csd_time,csd_parts,csd_selected,schema_ignore_params,validationSchemaIgnoreParams'
```
jfy133, line 22: Are these two really needed, line 26 seems to cover these already?

conf/cambridge.config
```diff
@@ -1,28 +1,80 @@
...
+}
+
+validation {
+    ignoreParams = ['partition', 'project', 'max_memory', 'max_cpus', 'max_time', 'csd_time', 'csd_parts', 'csd_selected', 'schema_ignore_params', 'validationSchemaIgnoreParams']
 }
 
-// Description is overwritten with user specific flags
+params.csd_time = {
```
jfy133, line 29: Can this not go in the main `params` block?

jfy133 (commented):
Sorry for any duplicate comments!

Don't forget to add the config in: https://github.com/nf-core/configs/blob/master/.github/workflows/main.yml

conf/cambridge.config
RaqManzano: We could but this is to be more compatible with older pipelines that still look for `schema_ignore_params` and `validationSchemaIgnoreParams`

conf/cambridge.config
RaqManzano: I prefer to keep it separate from the main block as it is not really a param for the users.

tavareshugo:
Thanks for this Raquel! A few comments and suggestions from my side: 

---

**`cambridge.md`**

In the guidelines, I would give a specific recommendation for where to store the singularity cache. 

In our HPC, it is **definitely not recommended** to store directly in `$HOME/` -- perhaps an explicit warning about this is good to include. 

You could also give a specific recommendation, for example: `$HOME/rds/hpc-work/nxf-singularity-cache` is a good place. 

---

I like your approach of defining the max parameters from the partition. At the risk of over-complicating, I wonder if there could be an "automatic" selection if the user doesn't specify anything. For example, in my config I have the following:

```groovy
process {
  executor = 'slurm'
  clusterOptions = --partition icelake'

  // Settings below are for CSD3 nodes detailed at
  //   https://docs.hpc.cam.ac.uk/hpc/index.html
  // Current resources (Jun 2023):
  //   icelake: 76 CPUs; 3380 MiB per cpu; 6760 MiB per cpu (himem)
  //   cclake: 56 CPUs; 3420 MiB per cpu; 6840 MiB per cpu (himem)
  // The values used below were chosen to be multiples of these resources
  // assuming a maximum of 2 retries

  // Using himem partition to ensure enough memory for single-CPU jobs
  withLabel:process_single {
      cpus   = { check_max( 1                  , 'cpus'    ) }
      memory = { check_max( 6800MB * task.attempt, 'memory'  ) }
      time   = { check_max( 4.h  * task.attempt, 'time'    ) }
      clusterOptions = "--partition icelake-himem"
  }
  // 4 CPUs + 13GB RAM
  withLabel:process_low {
      cpus   = { check_max( 4     * task.attempt, 'cpus'    ) }
      memory = { check_max( 13.GB * task.attempt, 'memory'  ) }
      time   = { check_max( 4.h   * task.attempt, 'time'    ) }
      clusterOptions = "--partition icelake"
  }
  // 8 CPUs + 27GB RAM
  withLabel:process_medium {
      cpus   = { check_max( 8     * task.attempt, 'cpus'    ) }
      memory = { check_max( 27.GB * task.attempt, 'memory'  ) }
      time   = { check_max( 8.h   * task.attempt, 'time'    ) }
      clusterOptions = "--partition icelake"
  }
  // 12 CPUs + 40GB RAM
  withLabel:process_high {
      cpus   = { check_max( 12    * task.attempt, 'cpus'    ) }
      memory = { check_max( 40.GB * task.attempt, 'memory')}
      time   = { check_max( 8.h  * task.attempt, 'time'    ) }
      clusterOptions = "--partition icelake"
  }
  // Going by chunks of 12h (2 retries should bring it to max of 36h)
  withLabel:process_long {
      time   = { check_max( 12.h  * task.attempt, 'time'    ) }
  }
  // A multiple of 3 should bring it to max resources on icelake-himem
  withLabel:process_high_memory {
      cpus   = { check_max( 25     * task.attempt, 'cpus'    ) }
      memory = { check_max( 170.GB * task.attempt, 'memory' ) }
      clusterOptions = "--partition icelake-himem"
  }
  withLabel:error_ignore {
      errorStrategy = 'ignore'
  }
  withLabel:error_retry {
      errorStrategy = 'retry'
      maxRetries    = 2
  }
}
```

So, I allow for  2 retries and increase the resources accordingly, to roughly reach the maximum resources of each partition. 
So, the behaviour could be something like: 

- If user specifies `--partition`, then all jobs are submitted to that partition only
- If user doesn't specify anything, something like the above would kick in, where jobs are submitted either to `icelake` or `icelake-himem` depending on the process labels.

---

But, totally fine if you prefer to leave these extra additions out - the config revision you did is already a great update to the previous config, thanks so much! 🙂

RaqManzano:
Thanks for the comments @tavareshugo, I just changed $HOME to $HOME/rds, you were right. it was lazy writing from my side, thanks for calling on that! Regarding the "automatic" selection scenario, I already tried an implementation of it and after discussing with the nf-core guys we went for a simpler approach. I refined the docs with info about the partitions. Let me know if you agree, feel free to make changes.

docs/cambridge.md
```diff
@@ -1,87 +1,131 @@
...
+screen -S nextflow
+
+# Re-attach later if needed
+tmux attach -t nextflow
+screen -r nextflow
+```
+
+Detaching from `tmux` leaves the workflow running in the background with
+`Ctrl-b` then `d`. For `screen`, use `Ctrl-a` then `d`.
+
```
tavareshugo, line 117:
````suggestion

You can then logout of the HPC and reattach to the session later. 
Before logging out, make sure to **note the node you’re on**. 
Suppose your login node was called `login-p-3`, you can later log back into this specific node as follows:

```bash
ssh username@login-p-3.hpc.cam.ac.uk
```

Then, you can re-attach to the `tmux`/`screen` session:

```bash
tmux attach -t nextflow
screen -r nextflow
```
````

File: docs/cambridge.md [hunk 5]
```diff
@@ -1,87 +1,131 @@
...
+session so that the Nextflow manager process keeps running after you disconnect
+your SSH session.
+
 ```
+# Start a tmux session
+tmux new -s nextflow
+
+# Or start a screen session
+screen -S nextflow
+
```
tavareshugo, line 109: I guess you can remove the reattach commands, if you accept my suggestion below.

tavareshugo (commented):
Keeping the config "simpler" is probably a good idea indeed. 

I've made a couple of minor suggestions to the markdown.

PR #1108, opened 2026-05-12
Fix the --mem request not allowed error when running on Trillium nodes

joshuayangca:
@MareikeJaniak 
@CMonnin 

Fix SBATCH ERROR
"The --mem=... request is not allowed nor necessary on Trillium; all nodes have the same amount of available memory (745 GiB) and each job get all the available memory of the node"

### DIFF
--- a/conf/alliance_canada.config
+++ b/conf/alliance_canada.config
@@ -47,6 +47,12 @@ process {
     cpus            = 1
     time            = '1h'
 
+
+  withLabel: 'process_single' { memory = null }
+  withLabel: 'process_low'    { memory = null }
+  withLabel: 'process_medium' { memory = null }
+  withLabel: 'process_high'   { memory = null }
+
 // NOTE:
 // these resourceLimits are set to baseline CPU for each cluser
 // Currently missing are configs for GPUs


pontus (approved):

pontus:
While I don't know anything in particular about this site, a fair amount of clusters that do not allow shared-node jobs still thinks nodes should be used fully, which I suspect won't be the case for many pipelines.

Since the nodes seem large enough to support it, it may be a good idea to add a note that using the `local` executor on a node may be a better option.

pontus:
One drawback with this solution is that while it lets a lot of processes run, many pipelines also either use other labels or use `withName` clauses to assign resources for specific processes, both of which will probably break here.

CMonnin (changes requested):
these changes seem to be unconditional. they will apply to all the clusters.

CMonnin:
just an fyi the use of trillium with normal nf-core pipelines is discouraged. as per the wiki.
https://docs.alliancecan.ca/wiki/Nextflow

> We discourage you from running nf-core pipelines or any other generic Nextflow pipeline on Trillium. We recommend running a pipeline on Trillium only if it was designed specifically for Trillium.

PR #1132, opened 2026-06-17
[WIP] nf-co2footprint activation on CFC cluster

JosuaCarl:
---
name: CFC nf-co2footprint activation
about: Activation of the nf-co2footprint plugin per default
---


JosuaCarl:
Hi, I was wondering, why there is a check that prevents the activation of a plugin? We are sure that we would like to activate `nf-co2footprint` per default on our machines, so I don't see why this should be prevented.

PR #1136, opened 2026-07-16
Add Kelvin2 (QUB) institutional config

wmsci:
## Summary

Add an institutional config profile for Queen's University Belfast's Kelvin2 HPC cluster (`-profile kelvin2`). The profile has been successfully tested on the Kelvin2 cluster.

## Steps for adding a new config profile:

### DIFF
--- a/conf/kelvin2.config
+++ b/conf/kelvin2.config
@@ -61,7 +61,6 @@ process {
 executor {
     queueSize       = 200
     submitRateLimit = '10/sec'
-    exitReadTimeout = '30 min'   // Wait longer for the .exitcode file on a busy scheduler.
     jobName         = { "${task.process.split(':').last()}" }
 }
 
--- a/conf/kelvin2.config
+++ b/conf/kelvin2.config
@@ -1,0 +1,82 @@
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Nextflow config for QUB's Kelvin2 HPC
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Author: William McCann
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+*/
+
+params {
+    config_profile_name        = 'kelvin2'
+    config_profile_description = "Queen's University Belfast Kelvin2 HPC cluster profile provided by nf-core/configs."
+    config_profile_contact     = 'William McCann (@wmsci)'
+    config_profile_url         = 'https://ni-hpc.github.io/nihpc-documentation/'
+
+    // Legacy resource ceilings. Only read by older nf-core pipelines (pre-v3.0 template
+    // or Nextflow < 24.04). Modern pipelines use `process.resourceLimits` below instead.
+    max_cpus   = 128
+    max_memory = 2.TB
+    max_time   = 720.h
+}
+
+process {
+    // Modern resource ceilings, used by current Nextflow and nf-core pipelines.
+    // Mirrors params.max_* above so both old and new pipelines are covered.
+    resourceLimits = [
+        cpus  : params.max_cpus,
+        memory: params.max_memory,
+        time  : params.max_time,
+    ]
+
+    executor   = 'slurm'
+    maxRetries = 2
+
+    // Timestamp-tolerant caching
+    cache      = 'lenient'
+
+    // Dynamic partition (queue) selection. Nextflow sets the SLURM partition (-p) from the
+    // return value and the walltime (-t) from task.time.
+    //   - Jobs above the 786 GB general-node ceiling go only to k2-himem.
+    //   - Other CPU tasks are placed by walltime; high memory-per-core jobs are additionally offered k2-himem.
+    queue = {
+        // General nodes top out at 786 GB; only k2-himem (2 TB, 3-day limit) can take more.
+        if (task.memory && task.memory > 786.GB) {
+            return 'k2-himem'
+        }
+        // Offer eligible partitions from highest to lowest priority: k2-hipri (≤ 3 h) and
+        // k2-medpri (≤ 24 h) only within their walltime limits, then k2-lowpri as a catch-all.
+        def partitions = []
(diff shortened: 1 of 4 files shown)


conf/kelvin2.config
```diff
@@ -0,0 +1,82 @@
...
+            partitions.add('k2-himem')
+        }
+        return partitions.join(',')
+    }
+}
+
+executor {
+    queueSize       = 200
+    submitRateLimit = '10/sec'
+    exitReadTimeout = '30 min'   // Wait longer for the .exitcode file on a busy scheduler.
```
pontus: Since the comment about the scheduler is here - this is rather used to control a file system thing. Since this is the kind of things that shouldn't happen, having a longer timeout here mostly means one will have to wait a rather long time for jobs that do not exit "properly" - e.g. either time out or get OOM-killed in a way that they don't get an `.exitcode` file.

(I have some slight memories of this being used to work around scheduler issues a long time ago, but nextflow has become much better at dealing with those and won't fail the same way, so if that was the case, I still think it's moot now).

docs/kelvin2.md
```diff
@@ -0,0 +1,70 @@
+# nf-core/configs: Queen's University Belfast Kelvin2 HPC Configuration
+
+nf-core pipelines have been configured for use on Queen's University Belfast's [Kelvin2](https://ni-hpc.github.io/nihpc-documentation/) HPC cluster. To use this profile, run a pipeline with `-profile kelvin2`, which loads the pre-configured [`kelvin2.config`](../conf/kelvin2.config).
+
+This profile was tested with Nextflow `26.04.4` and the `nf-core/rnaseq` pipeline `3.26.0`.
+
+## Set up Nextflow and Apptainer
+
+Nextflow and Apptainer can both be installed into a Conda or Mamba environment.
```
pontus: Is neither apptainer or nextflow available natively?

I personally have rather strong views about avoiding conda on HPC because it makes things worse for everyone (apart from the space wasted, inodes/files is a commodity that is also precious and in particular visible with how difficult it is to do backups sensibly where there is a lot of conda use).

Nextflow is very much just a wrapper script that does everything, so I'd strongly recommend a `wget`/`curl` to acquire that to a good location (although not the piping through to a shell variant).

I haven't checked recently, but the apptainer in conda used to be built so it opted to unpack images rather than mount them. That's inefficent time wise and requires a potential lot of temporary space that might be what's stopping jobs from completing because they fill up some file system. There has also been quite a few cases where this breaks because of file attributes that aren't an issue when using an apptainer that uses FUSE to mount images instead.

If the system doesn't provide apptainer installed at a system service, I'd strongly suggest requesting it (and if it has one but it doesn't mount images using FUSE but rather unpacks them, that could be worth asking about as well).

docs/kelvin2.md
```diff
@@ -0,0 +1,70 @@
...
+
+The Nextflow driver process must keep running for the whole pipeline. Two recommendations:
+
+- **Use a terminal multiplexer such as `tmux` or `screen`** so the driver survives an SSH disconnect. Note which login node you are on so you can reconnect to the same one.
+
+  ```bash
+  tmux new -s nextflow        # detach: Ctrl-b then d; reattach: tmux attach -t nextflow
+  ```
+
+- **Do not run heavy work on a login node.** For anything beyond a quick test with pre-cached container images, start the driver inside an interactive job (it will submit the pipeline's own jobs from there):
```
pontus: Local policies should of course be adhered to, but the nextflow main monitoring process is decidedly not heavy for most pipelines (with the possible exception of if it needs to run apptainer/singularity to pull images).

I have at least successfully advocated for policies at sites that make it explicitly okay to run this kind of thing on a login/submission node rather than wasting compute resources, but again, local policies should be reflected in advice.

docs/kelvin2.md
```diff
@@ -0,0 +1,70 @@
...
+
+  ```bash
+  srun --partition=k2-medpri --ntasks=1 --mem-per-cpu=4G --time=24:00:00 --pty bash
```
pontus: If one does suggest getting a job, it might be more useful to have a sample job script for that - something that can be handled to `sbatch` that will run hands free?

conf/kelvin2.config
wmsci: Thanks. That makes sense. I've removed `exitReadTimeout`.

docs/kelvin2.md
wmsci: Thanks for the detailed notes. Kelvin2 provides both as modules, so I updated the docs to recommend those first. I've also added an option to install a newer Nextflow binary onto scratch with curl (written to a file, not piped to a shell) as you recommended. I've kept Conda as an option as it isn't discouraged by Kelvin2 support staff, but it's no longer the primary advice.

docs/kelvin2.md
wmsci: Agreed, I've softened the advice here. The docs now say the main nextflow process is fine on a login node once images are cached, and only suggest a compute node for first-time / uncached image pulls. Personally, I've seen Apptainer pulls hit CPU harder than I'd like on Kelvin2 login nodes, so this is intentionally conservative advice from me to help people avoid getting told off for heavy login node use.

docs/kelvin2.md
wmsci: I've added a sample `sbatch` script for that case. Thank you for thoughtful review and the helpful suggestions.

docs/kelvin2.md
```diff
@@ -0,0 +1,110 @@
...
+This profile was tested with Nextflow `26.04.4` and the `nf-core/rnaseq` pipeline `3.26.0`.
+
+## Set up Nextflow and Apptainer
+
+### Option 1: Environment modules (recommended)
+
+Kelvin2 provides Nextflow and Apptainer as modules:
+
+```bash
+module load nextflow/24.07.27/java-22.0.2
```
pontus, line 14: Since the module here implies a rather old install, it might be nice mentioning that you can normally run different versions of nextflow without replacing your `nextflow` wrapper, e.g.

`NXF_VER=26.04.6 nextflow run nextflow-io/hello`

pontus (approved):

docs/kelvin2.md
wmsci: Thanks, added this.

PR #1138, opened 2026-07-17
Workaround for FCS-GX in Tree of Life pipelines at the Sanger

muffato:
We're unable to make FCS-GX run through Singularity. Our workaround is to use a bare-metal installation that is provided as a module.
Conda works too but cannot be considered a fallback for singularity because the singularity profile does `conda.enabled = false`

The `withName` rule is restricted to https://github.com/sanger-tol pipelines, which we control.

---
name: New Config
about: A new cluster config
---

### DIFF
--- a/conf/sanger.config
+++ b/conf/sanger.config
@@ -45,3 +45,16 @@ includeConfig ({
         return "/dev/null"
     }
 }.call())
+
+
+// HPC optimisations and workarounds
+process {
+
+    // From @muffato and @DLBPointon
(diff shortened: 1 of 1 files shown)


pontus (approved):

pontus:
Approving and guessing this might be the way that makes most sense in practice, but otherwise, it seems this would rather be a pipeline config?

muffato:
I put it here because the module will be used in multiple pipelines and I thought it'd be better to define the workaround just once.

PR #1140, opened 2026-07-19
SAGE: remove stale strict-syntax deprecation warning

adamjtaylor:
## Summary

The `sage` config's deprecation banner (added in #1089) warns that the config will break under Nextflow's upcoming strict config syntax. That is no longer the case — `conf/sage.config` already parses and runs cleanly under the strict parser. This PR removes the now-inaccurate banner and documents one remaining cosmetic warning.

### Verification

- `nextflow lint conf/sage.config` → ✅ no errors (Nextflow 26.04.6, strict parser)
- Strict-runtime smoke test: `nextflow run main.nf -c conf/sage.config` runs cleanly

### Changes

- **Remove the `warning_message = { System.out.println(...) }.call()` block.** The config is strict-syntax compatible, so the "will be deprecated by end of May 2026" message is inaccurate and prints to stdout on every run.
- **Keep `threadPool.FileTransfer.maxAwait = '24 hour'` and add an explanatory comment.** This key is still read and honoured at runtime by `FilePorter`/`ThreadPoolManager` (default is 12h; this doubles it for large S3 stagings). The `Unrecognized config option` message emitted by the config schema validator (>=25.04) is a known false positive because the `threadPool.*` scope is undocumented in the schema — the value is still applied. The comment prevents a future maintainer from "cleaning up" a working setting.

No functional/resource-allocation changes.

Refs #1020 (strict syntax migration).

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated (n/a — no changelog file for per-config changes)
- [x] The config has been tested with the strict Nextflow parser

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conf/sage.config
+++ b/conf/sage.config
@@ -14,24 +14,14 @@ params {
 
     // Define task exit errors
     exit_status_scaling = [143,137,104,134,139,247]
-
-    warning_message = {
-        System.out.println("WARNING: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
-        System.out.println("WARNING:")
-        System.out.println("WARNING: THIS CONFIG IS NO LONGER MAINTAINED.")
-        System.out.println("WARNING:")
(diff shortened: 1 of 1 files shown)


rxu17 (approved):
LGTM! Thanks for doing this!

pontus (approved):

PR #1143, opened 2026-07-21
Add verily_workbench profile

samhornstein:
## Add `verily_workbench` profile

Adds an institutional profile for running nf-core pipelines on [Verily Workbench](https://workbench.verily.com/) via the Google Cloud Batch executor.

The profile is a Workbench-specialized sibling of `googlebatch`: it reads project and worker service account from the environment variables Workbench injects into each job (`GOOGLE_CLOUD_PROJECT`, `GOOGLE_SERVICE_ACCOUNT_EMAIL`) and defaults to the VPC network layout Workbench provisions in every workspace, while keeping all values overridable as params.

Tested end-to-end on Google Cloud Batch from a Verily Workbench workspace with `nf-core/demo` and `nf-core/rnaseq` (`-profile test,verily_workbench`); both completed successfully.

### Steps for adding a new config profile

- [x] OPTIONAL: Add your custom profile path and GitHub user name to `.github/CODEOWNERS`

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,3 +15,4 @@
 **/mahuika** @jen-reeve
 **/mcw_rcc** @semenko
 **/purdue_** @aseetharam
+**/verily_workbench** @samhornstein
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -211,6 +211,7 @@ jobs:
           - "uzh"
           - "uzl_omics"
           - "vai"
+          - "verily_workbench"
           - "vsc_calcua"
           - "vsc_kul_uhasselt"
           - "vsc_ugent"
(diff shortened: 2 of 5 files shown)


pontus:
Looks simple enough, there's no need to set and `resourceLimits` or similar?

Approving now, but it might be nice to include the `params` you introduce in an ignore list for nf-schema so pipelines don't complain about them.

(I carelessly clicked rebase that might make it slightly more cumbersome to update, feel free to force-push if that's easier for updates.)

pontus (approved):

samhornstein:
> Looks simple enough, there's no need to set and `resourceLimits` or similar?
> 
> Approving now, but it might be nice to include the `params` you introduce in an ignore list for nf-schema so pipelines don't complain about them.
> 
> (I carelessly clicked rebase that might make it slightly more cumbersome to update, feel free to force-push if that's easier for updates.)

Thanks @pontus! Added `validation.ignoreParams` based on your recommendation. Intentionally left off `resourceLimits`; the [official googlebatch.config](https://github.com/nf-core/configs/blob/master/conf/googlebatch.config) doesn't set it either.

PR #1153, opened 2026-07-29
Fix apptainer access to node-local scratch

manascripts:
Apptainer containers currently only bind `/omics`. Minor change to bind `/local` globally in the profile so that LSF-provided local scratch remains accessible to all containerized tasks.


manascripts:
@abidabrar, I was facing this issue with some tools - do you agree with this change?

pontus (approved):

PR #1154, opened 2026-08-03
Update sanger.config

DLBPointon:
Removal of the FCSGX_RUNGX config addition

We forgot this meant people would get warnings about tools they arn't using. Instead we'll patch the profile locally for effected pipelines.

### DIFF
--- a/conf/sanger.config
+++ b/conf/sanger.config
@@ -46,15 +46,3 @@ includeConfig ({
     }
 }.call())
 
-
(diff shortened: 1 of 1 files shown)


jfy133 (commented):
Why not make a pipeline specific config , so it only applies to pipelines that has the tool?

pontus (approved):

prototaxites (approved):

muffato (approved):

PR #1160, opened 2026-08-10
Add GPU configuration to Pelle (UPPMAX) profile

CormacKinsella:
---
name: GPU configuration on pelle
about: Configures SLURM resources for GPU labelled processes
---

- For GPU labelled processes, the SLURM partition & requested resources must be configured on the workflow side, which is not reproducible across clusters
- The PR adds configuration for the Pelle cluster and has been tested using `parabricks fq2bam`

### DIFF
--- a/conf/uppmax/pelle.config
+++ b/conf/uppmax/pelle.config
@@ -12,6 +12,7 @@ params {
 
 // Hardware: https://docs.uppmax.uu.se/hardware/clusters/pelle/#pellemaja-hardware
 // Resource limits: https://docs.uppmax.uu.se/cluster_guides/slurm_on_pelle/#the-pelle-partition
+// GPU modules: https://nf-co.re/docs/developing/components/gpu-modules
 process {
     resourceLimits = [
         cpus: 96,
--- a/conf/uppmax/pelle.config
+++ b/conf/uppmax/pelle.config
@@ -22,20 +22,12 @@ process {
         [
             "-A ${params.project}",
             params.clusterOptions ?: "",
-            task.memory > 768.GB
-                ? "-p fat -C ${task.memory > 2.TB ? '3TB' : '2TB'}"
-                : "",
+            task.accelerator
+                ? "-p ${params.gpu_partition} --gpus=${params.gpu_gres}"
+                : task.memory > 768.GB
+                    ? "-p fat -C ${task.memory > 2.TB ? '3TB' : '2TB'}"
+                    : "",
         ].minus("").join(" ")
-    }
-    withLabel: process_gpu {
-        clusterOptions = {
-            [
-                "-A ${params.project}",
-                params.clusterOptions ?: "",
-                "-p ${params.gpu_partition}",
-                "--gpus=${params.gpu_gres}"
-            ].minus("").join(" ")
-        }
     }
 }
 
--- a/conf/uppmax/pelle.config
+++ b/conf/uppmax/pelle.config
@@ -4,7 +4,6 @@
 params {
     config_profile_description = 'UPPMAX (Pelle) cluster profile provided by nf-core/configs.'
     gpu_gres                   = 'l40s:1' // GPU generic resources: https://slurm.schedmd.com/gres.html
-    gpu_partition              = 'gpu'
     max_cpus                   = 96
     max_memory                 = 3.TB
     max_time                   = 10.d
@@ -12,7 +11,6 @@ params {
 
 // Hardware: https://docs.uppmax.uu.se/hardware/clusters/pelle/#pellemaja-hardware
 // Resource limits: https://docs.uppmax.uu.se/cluster_guides/slurm_on_pelle/#the-pelle-partition
-// GPU modules: https://nf-co.re/docs/developing/components/gpu-modules
 process {
     resourceLimits = [
         cpus: 96,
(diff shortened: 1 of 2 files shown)


conf/uppmax/pelle.config
```diff
@@ -6,6 +6,8 @@ params {
     max_cpus                   = 96
     max_memory                 = 3.TB
     max_time                   = 10.d
+    gpu_partition              = 'gpu'
+    gpu_gres                   = null
```
mahesh-panchal: what does `gres` mean?

conf/uppmax/pelle.config
```diff
@@ -25,6 +27,16 @@ process {
                 : "",
         ].minus("").join(" ")
     }
+    withLabel: process_gpu {
+        clusterOptions = {
+            [
+                "-A ${params.project}",
+                params.clusterOptions ?: "",
+                "-p ${params.gpu_partition}",
```
mahesh-panchal: There's only the one partition right? So we don't need an extra parameter for this.

conf/uppmax/pelle.config
pontus: `gres` is slurm-speak for generic resource, something that can be used for scheduling decisions.

CormacKinsella: It comes from older SLURM versions where you'd request a GPU with e.g: `--gres=gpu:1`, it means generic resources (https://slurm.schedmd.com/gres.html).

Want to go for `gpu_resources` or something more self-explanatory?

pontus: Any reason not putting the default here as for other `params`?

CormacKinsella: Will fix that, thanks

pontus: This would likely complain about the unknown `params`. Add new `params` to the ignore list so nf-schema doesn't complain.

conf/uppmax/pelle.config
```diff
@@ -25,6 +27,16 @@ process {
                 : "",
         ].minus("").join(" ")
     }
+    withLabel: process_gpu {
```
pontus: What we say for nf-core developers is to make sure `accelerator` is set, and this could be included in the "normal" `clusterOptions` looking at `task.accelerator` to determine whatever to amend the request with anything about GPUs.

pontus: (https://nf-co.re/docs/developing/components/gpu-modules#detecting-a-gpu-request)

conf/uppmax/pelle.config
mahesh-panchal: I think put a short comment then afterwords spelling it out, and a link to those docs.
```suggestion
    gpu_gres                   = 'l40s:1' // GPU generic resources: https://slurm.schedmd.com/gres.html
```
If anyone is used to setting that flag, it's better to keep the UI as similar as possible.

conf/uppmax/pelle.config
mahesh-panchal: Ah, so that's why it's set in the base.config:
```
        ext.use_gpu = { workflow.profile.contains('gpu') }
        accelerator = { workflow.profile.contains('gpu') ? 1 : null }
```
So these need to be added here.

conf/uppmax/pelle.config
mahesh-panchal: scratch that. There's also the `haswell` partition, with the T4's.

conf/uppmax/pelle.config
CormacKinsella: @mahesh-panchal - I might misunderstand but I don't think these need to be restated. Instead we can remove the withLabel block, and just infer GPU via task.accelerator which should have been set either via `withName` for e.g. `ribodetector`
https://github.com/nf-core/modules/blob/master/modules/nf-core/ribodetector/main.nf
or `withLabel` for e.g. parabricks.

pontus (approved):
Haven't verified but looks functional.

conf/uppmax/pelle.config
```diff
@@ -20,9 +22,11 @@ process {
         [
             "-A ${params.project}",
             params.clusterOptions ?: "",
-            task.memory > 768.GB
-                ? "-p fat -C ${task.memory > 2.TB ? '3TB' : '2TB'}"
-                : "",
+            task.accelerator
+                ? "-p ${params.gpu_partition} --gpus=${params.gpu_gres}"
```
mahesh-panchal: I was thinking we should include an map here instead of two parameters,
and a link to the docs.
```
// https://docs.uppmax.uu.se/cluster_guides/slurm_on_pelle/#partitions-on-pelle
def gpu_partition = [
    'l40s': 'gpu',
    'h100': 'gpu',
    't4': 'haswell'
]
```
It makes it easy to see at a glance what's available. We can error check it too. And it's simple to keep up to date.
So then this string is just formed `gpu_gres` parameter instead.

conf/uppmax/pelle.config
mahesh-panchal: I misunderstood. The settings don't need to be included again here since we're not overriding the `process_gpu` label. 

Although I understood `task.accelerator` only has a value if `-profile gpu` is used ( and without setting any other `withLabel`/`withName` stuff ourselves). This then has the value 1 generally, but actual accelerator settings from the institutional configs which override it when the executor supports it like aws-batch, etc. 

I think it would be useful to comment a link to the docs showing checking `task.accelerator` is the appropriate method here. It's useful to anyone else that goes reading these things which is fairly often.

mahesh-panchal (approved):
Thank you. LGTM 👍🏽

conf/uppmax/pelle.config
```diff
@@ -17,12 +18,28 @@ process {
...
+
         [
             "-A ${params.project}",
             params.clusterOptions ?: "",
-            task.memory > 768.GB
-                ? "-p fat -C ${task.memory > 2.TB ? '3TB' : '2TB'}"
-                : "",
+            task.accelerator // Follows implementation pattern described by https://nf-co.re/docs/developing/components/gpu-modules
+                ? "-p ${gpu_partition[gpu_type]} --gpus=${params.gpu_gres}"
+                : task.memory > 768.GB
```
mahesh-panchal, line 40: One thing I just realized too late. This pattern does not let someone use the gpu profile and submit to the high memory cpu node.

CormacKinsella: I think it's safe, accelerator should only be set for gpu processes.

```
    withLabel: process_gpu {
        ext.use_gpu = { workflow.profile.contains('gpu') }
        accelerator = { workflow.profile.contains('gpu') ? 1 : null }
    }
```

mahesh-panchal: Right. Thanks for checking.

PR #1161, opened 2026-08-10
Remove deprecated pipeline-specific specific Maestro config

jfy133:
This is the last of the strict-syntax compliance campaign!

Already communicated with the maintainers who are moving to a local copy.


nvnieuwk (approved):

jfy133:
Well, we still need to remove UTD_Juno... but at least tests passing!

PR #1165, opened 2026-08-13
Replace params.gpu_gres with per-process accelerator on Pelle

mahesh-panchal:
## Summary

- Removes the pipeline-wide `params.gpu_gres` param from `conf/uppmax/pelle.config` in favor of Nextflow's standard `process.accelerator` directive, read per-task in `clusterOptions`.
- Adds a default of `[type: 'l40s', request: 1]` on any process labelled `process_gpu`, applied only when `-profile gpu` is active (`workflow.profile.contains('gpu')`) — matching the same gating nf-core's own `base.config` uses.
- Removes the now-dead `gpu_gres` entries from `conf/uppmax.config`'s ignore-params lists.
- Adds a "Using GPUs on Pelle" section to `docs/uppmax.md` documenting the per-process override pattern (different GPU type/count per process, and disabling the GPU for one process with `accelerator = null`).

## Why

A single `gpu_gres` param can't express different GPU needs for different processes in the same pipeline run, and changing it risks invalidating `-resume` state for unrelated tasks. Follows the design discussed in [NBISweden/biovat#45](https://github.com/NBISweden/biovat/issues/45#issuecomment-5277374410).

### DIFF
--- a/conf/uppmax/pelle.config
+++ b/conf/uppmax/pelle.config
@@ -3,7 +3,6 @@
 
 params {
     config_profile_description = 'UPPMAX (Pelle) cluster profile provided by nf-core/configs.'
-    gpu_gres                   = 'l40s:1' // GPU generic resources: https://slurm.schedmd.com/gres.html
     max_cpus                   = 96
     max_memory                 = 3.TB
     max_time                   = 10.d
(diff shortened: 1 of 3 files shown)


conf/uppmax/pelle.config
```diff
@@ -17,29 +16,37 @@ process {
...
-            )
+        def gpu_directive = ""
+        // Follows the nf-core GPU guidelines: https://nf-co.re/docs/developing/components/gpu-modules
+        if (task.accelerator) {
+            // https://docs.uppmax.uu.se/cluster_guides/slurm_on_pelle/#partitions-on-pelle
+            def gpu_partition = [
+                'l40s': 'gpu',
+                'h100': 'gpu',
+                't4': 'haswell'
+            ]
```
pontus, line 32: Wouldn't it make more sense to put this table in `params`?

pontus: Sorry, seems I was confusing this with another config that moved everything there, this can be disregarded.

conf/uppmax/pelle.config
```diff
@@ -17,29 +16,37 @@ process {
...
         time: 10.d,
     ]
+
+    // Default GPU for any process labelled `process_gpu` (e.g. via `-profile gpu`).
+    // Override per process in your own config, e.g.:
+    //   withName: 'GPU_PROCESS' { accelerator = [ type: 'h100', request: 2 ] }
+    withLabel: process_gpu {
+        accelerator = { workflow.profile.contains('gpu') ? [ type: 'l40s', request: 1 ] : null }
+    }
+
```
pontus: My reading of the GPU guidelines doesn't think this should be done by the institutional profile.

(It good to have examples in docs to have examples of how users can request specifics for a particular process.)

pontus: As a follow from that, I think I would prefer if the profile had a sensible default for when no `type` is specified, but also allowed for easy passing if users wanted something else. This means I think it makes slightly more sense to have the default gpu type in `params` (to be overridden "easily" if desired), while still allowing for more specific configurations if needed.

conf/uppmax/pelle.config
```diff
@@ -17,29 +16,37 @@ process {
         memory: 3.TB,
         time: 10.d,
     ]
+
+    // Default GPU for any process labelled `process_gpu` (e.g. via `-profile gpu`).
+    // Override per process in your own config, e.g.:
+    //   withName: 'GPU_PROCESS' { accelerator = [ type: 'h100', request: 2 ] }
+    withLabel: process_gpu {
```
CormacKinsella: Rather than set accelerator, can we not check and apply `task.accelerator.type`, reverting to 'l40s' if not found.

conf/uppmax/pelle.config
mahesh-panchal: The assignment to `gpu_type`, and the `process_gpu` override both provide defaults for `accelerator.type`.

I asked about what the best practices for gpu modules in https://nfcore.slack.com/archives/C07PPQR415W/p1786550303862009 but I haven't received a reply yet. 

However, to me, it makes sense that an institutional config sets this and a default type as it is what defines the hardware available. 

I don't think a pipeline should define these parameters because they should be agnostic of the hardware, and the most they could request is the gpu count.

conf/uppmax/pelle.config
mahesh-panchal: What difference does it make to the proposed behaviour?

conf/uppmax/pelle.config
pontus: I suspect we think the same but have different ideas about how that should happen.

I agree the institutional profile should set the type. That doesn't mean it should have a `process_gpu` overriding any `accelerator` set by the pipeline (I suspect it's a rare occurrence so more a theoretical issue, but assuming software that requires multiple cards that the pipeline sets with `accelerator = 4`, that shouldn't be overridden by this profile.

And I don't see the benefit of having this either, since we still need to handle cases where the pipeline (or user) sets in on other basis (e.g. `withName`).

mahesh-panchal: > And I don't see the benefit of having this either, since we still need to handle cases where the pipeline (or user) sets in on other basis (e.g. `withName`).

The fallback is there:
```
def gpu_type = task.accelerator.type ?: 'l40s'
```
It'll get `l40s` if type is not defined, so even if someone doesn't follow the readme examples and just sets a number, it will still be correctly set with the default.

The `process_gpu` override becomes mostly a visual thing to define the default in one place.

pontus: I think that's my hangup here, with the `process_gpu` not doing anything (since it can't be relied upon), potentially breaking intended behaviour (although again, mostly theoretical) and overall not being how I think this should work.

conf/uppmax/pelle.config
CormacKinsella: It leaves configuration of accelerator to the pipeline in line with nf-core docs, but still allows users to override specific process resources with -c since it can accept the map. 

Caveat, not tested it, just thinking.

conf/uppmax/pelle.config
mahesh-panchal: Dropped the `withLabel: process_gpu` override in 9534b07f. The `clusterOptions` fallback (`task.accelerator.type ?: ...`) already covers the same default behaviour.

For the "easily overridable default type" point — added `params.pelle_gpu_type` (default `l40s`), so `--pelle_gpu_type h100` now works pipeline-wide, while any per-process `accelerator` (via `withName` or the pipeline itself) still takes precedence.

pontus (approved):

conf/uppmax/pelle.config
CormacKinsella, line 54:
```suggestion
    runOptions = '-B $SNIC_TMP --nv'
```

CormacKinsella (changes requested):
`gpu` profile sets `--nv` but it's overridden by institutional.

If this needs to run without local config, we need to make the GPUs available.

CormacKinsella (approved):

PR #1170, opened 2026-08-27
Add local executor resource limits to Pelle profile

mahesh-panchal:
## Summary

- Cap the `local` executor on the `pelle` profile to 8 CPUs / 128GB memory, so processes marked to run locally (short jobs that don't go through Slurm) don't fan out and overburden the shared Pelle login node.

### DIFF
--- a/conf/uppmax/pelle.config
+++ b/conf/uppmax/pelle.config
@@ -55,3 +55,11 @@ singularity {
     // Forward Slurm's per-job GPU visibility restriction into the container.
     envWhitelist = 'SNIC_TMP,CUDA_VISIBLE_DEVICES'
 }
+
+// Cap local jobs so they don't overburden the shared login node
+executor {
+    $local {
+        cpus   = 8
+        memory = 128.GB
+    }
+}


pontus (approved):

pontus:
I thought I had approved this, sorry for the delay.

I think this is valid, but am a bit unsure about the actual effects - for a process marked with the local executor, will it just cause the execution to blow up if it requests more than these limits? Is there any better way to deal with it in that case (I can't easily think of one).

mahesh-panchal:
No, it doesn't blow up. It just queues them up locally, and runs a new task when enough resources are available.

pontus:
But if the requests are above the limits, enough resources will never become available?

pontus:
So I assume there should be a

```
Caused by:
  Process requirement exceeds available CPUs -- req: 16; avail: 8
```
in those cases?

mahesh-panchal:
These jobs should also be few and far between. I don't think any nf-core pipeline explicitly sets local executor on a process. This should only be set by short running jobs which shouldn't be submitted to the queue.

Ah, I get the question now. Sorry, `nextflow` will error and say there's not enough resources, but no one should be requesting this on the login node. Those kind of jobs should be going to `slurm`.

pontus:
Yes, I think it's acceptable to leave this to the user to deal with (they'd be the ones who'd be doing this anyway), just wanted to check if others thought there was anything smart that could be done.

mahesh-panchal:
If they're using this profile and this happens, it means they've likely explicitly set the `local` executor for a process, and they should know how update it.

PR #1174, opened 2026-09-03
naiss profile, including support for arrhenius

pontus:
---
name: NAISS
about: Profile to support naiss resources, starting with arrhenius hpc
---

This is work-in-progress, please do not merge until confirmed working.

### DIFF
--- a/conf/naiss.config
+++ b/conf/naiss.config
@@ -1,0 +1,90 @@
+// NAISS Config Profile
+// Docs: https://github.com/nf-core/configs/blob/master/docs/naiss.md
+// Supported clusters:
+// - Arrhenius
+
+
+params {
+    // Description is overwritten for other clusters using includeConfig
+    config_profile_description   = 'NAISS resource profile provided by nf-core/configs.'
+    config_profile_contact       = 'Pontus Freyhult (@pontus)'
+    config_profile_url           = 'https://www.naiss.se/'
+    project                      = null
+    clusterOptions               = null
+    schema_ignore_params         = "cluster-options,clusterOptions,project,naiss_gpu_type"
+    validationSchemaIgnoreParams = "cluster-options,clusterOptions,project,naiss_gpu_type,schema_ignore_params"
+    save_reference               = true
+   
+    // Fallback-defaults, these should be 
+    max_memory                   = 500.GB
+    max_cpus                     = 16
+    max_time                     = 240.h
+    // illumina iGenomes reference file paths on UPPMAX
+    ignore_params_list           = [
+        "cluster-options",
+        "clusterOptions",
+        "clusterName",
+        "ignore_params_list",
+        "max_cpus",
+        "max_memory",
+        "max_time",
+        "naiss_gpu_type",
+        "project",
+        "schema_ignore_params",
+        "validationSchemaIgnoreParams",
+    ]
+}
+
+// nf-schema/nf-validation settings
+// Different versions use different specifications
+validation {
+    ignoreParams = params.ignore_params_list
+}
+params.schema_ignore_params = params.ignore_params_list.join(",")
+params.validationSchemaIgnoreParams = params.ignore_params_list.join(",")
+
+apptainer {
+    enabled      = true
+    envWhitelist = 'NAISS_TMP,CUDA_VISIBLE_DEVICES'
+}
+
+executor {
+    $slurm {
+        account = params.project
+    }
+
+    // Cap local jobs to limit burden on the node they run on
+    $local {
+        cpus   = 8
+        memory = 128.GB
+    }
+}
+
+process {
+    executor       = 'slurm'
+    // Use node local storage for execution.
+    scratch        = '$NAISS_TMP'
+
+    resourceLimits = {
+        return [
+        cpus: params.max_cpus,
+        memory: params.max_memory,
+        time: params.max_time,
+    ]}
+
+}
+
+// Get additional settings dependent on resource, do not try to deal
+// with non-HPC resources right now
+includeConfig ({
+    def resource = "unknown"
+    try {
+        resource = ['/bin/bash', '-c', "sacctmgr show cluster -P -n | cut -f1 -d'|'"].execute().text.trim()
+    }
+    catch (IOException _e) {
+        System.err.println("WARNING: Could not run sacctmgr, defaulting to unknown")
+        return "/dev/null"
+        }
+    return resource ? "naiss/${resource}.conf": "/dev/null"
+   }.call())
+
--- a/conf/naiss.config
+++ b/conf/naiss.config
@@ -19,7 +19,6 @@ params {
     max_memory                   = 500.GB
     max_cpus                     = 16
     max_time                     = 240.h
-    // illumina iGenomes reference file paths on UPPMAX
     ignore_params_list           = [
         "cluster-options",
         "clusterOptions",
@@ -85,6 +84,6 @@ includeConfig ({
         System.err.println("WARNING: Could not run sacctmgr, defaulting to unknown")
         return "/dev/null"
         }
-    return resource ? "naiss/${resource}.conf": "/dev/null"
+    return resource ? "naiss/${resource}.config": "/dev/null"
    }.call())
 
--- a/conf/naiss.config
+++ b/conf/naiss.config
@@ -51,12 +51,6 @@ executor {
     $slurm {
         account = params.project
     }
-
-    // Cap local jobs to limit burden on the node they run on
-    $local {
-        cpus   = 8
-        memory = 128.GB
-    }
 }
 
 process {
--- a/conf/naiss/arrhenius.config
+++ b/conf/naiss/arrhenius.config
@@ -4,36 +4,28 @@
 
 params {
     config_profile_description = 'NAISS Arrhenius cluster profile provided by nf-core/configs.'
-    // gpus type is only for clarity in logs and such, arrhenius only has a
-    // single gpu type and we don't overcomplicate with code to break if one
-    // sets another
-    naiss_gpu_type             = 'GH200' 
     
     max_cpus                   = 192
     max_memory                 = 3.TB
     max_time                   = 3.d
 }
 
+// Some gymnastics to get conditional inclusion, here to cover cpu only
+includeConfig({
+    (params.account && params.account.trim().endsWith("-cpu")) ?
+        "arrhenius_cpu.config" :
+        "/dev/null"
+    }.call())
+
+// and GPU, we only do these for explicit account names
+includeConfig({
+    (params.account && params.account.trim().endsWith("-gpu")) ?
+        "arrhenius_gpu.config" :
+        "/dev/null"
+    }.call())
+
+
 // Info: https://www.naiss.se/resource/arrhenius/
-//
-
-// We only deal with cluster specific things here
-process {
-    clusterOptions = {
-        [
-            params.clusterOptions ?: "",
-
-            // Arrhenius only has one GPU type, and we don't implement logic
-            // just to give a different error. Also, see params above.
-
-            task.accelerator ? "-p gpu --gpus=${task.accelerator.request ?: 1}" : "",
-            task.memory > 768.GB ? "-p fat" : "",
-        ].minus("").join(" ")
-    }
-
-    // Enables GPU access inside the container when a task requests an accelerator.
-    containerOptions = { task.accelerator ? '--nv' : '' }
-}
 
 apptainer {
     runOptions = '-B $NAISS_TMP,/nobackup/proj/disk'
--- a/conf/naiss/arrhenius.config
+++ b/conf/naiss/arrhenius.config
@@ -1,0 +1,38 @@
+
+// Used by https://github.com/nf-core/configs/blob/master/conf/naiss.config
+// Cluster configuration for Arrhenius
+
+params {
+    config_profile_description = 'NAISS Arrhenius cluster profile provided by nf-core/configs.'
+    // gpus type is only for clarity in logs and such, arrhenius only has a
+    // single gpu type and we don't overcomplicate with code to break if one
+    // sets another
+    naiss_gpu_type             = 'GH200' 
+    
+    max_cpus                   = 192
+    max_memory                 = 3.TB
+    max_time                   = 3.d
+}
+
+// Info: https://www.naiss.se/resource/arrhenius/
+//
+
+// We only deal with cluster specific things here
+process {
+    clusterOptions = {
+        [
+
+            params.clusterOptions ?: "",
+            task.accelerator ? "-p gpu --gpus=${task.accelerator.request ?: 1}" : "",
+            task.memory > 768.GB ? "-p fat" : "",
+        ].minus("").join(" ")
+    }
+
+    // Enables GPU access inside the container when a task requests an accelerator.
+    containerOptions = { task.accelerator ? '--nv' : '' }
+}
+
+apptainer {
+    runOptions = '-B $NAISS_TMP,/nobackup/proj/disk'
+}
+
--- a/conf/naiss/arrhenius.config
+++ b/conf/naiss/arrhenius.config
@@ -21,8 +21,11 @@ params {
 process {
     clusterOptions = {
         [
-
             params.clusterOptions ?: "",
+
+            // Arrhenius only has one GPU type, and we don't implement logic
+            // just to give a different error. Also, see params above.
+
             task.accelerator ? "-p gpu --gpus=${task.accelerator.request ?: 1}" : "",
             task.memory > 768.GB ? "-p fat" : "",
         ].minus("").join(" ")
--- a/conf/naiss/arrhenius.config
+++ b/conf/naiss/arrhenius.config
@@ -12,14 +12,14 @@ params {
 
 // Some gymnastics to get conditional inclusion, here to cover cpu only
 includeConfig({
-    (params.account && params.account.trim().endsWith("-cpu")) ?
+    (params.project && params.project.trim().endsWith("-cpu")) ?
         "arrhenius_cpu.config" :
         "/dev/null"
     }.call())
 
 // and GPU, we only do these for explicit account names
 includeConfig({
-    (params.account && params.account.trim().endsWith("-gpu")) ?
+    (params.project && params.project.trim().endsWith("-gpu")) ?
         "arrhenius_gpu.config" :
         "/dev/null"
     }.call())
--- a/conf/naiss/arrhenius.config
+++ b/conf/naiss/arrhenius.config
@@ -5,7 +5,7 @@
 params {
     config_profile_description = 'NAISS Arrhenius cluster profile provided by nf-core/configs.'
     
-    max_cpus                   = 192
+    max_cpus                   = 256
     max_memory                 = 3.TB
     max_time                   = 3.d
 }
--- a/conf/naiss/arrhenius_cpu.config
+++ b/conf/naiss/arrhenius_cpu.config
@@ -1,0 +1,16 @@
+// Invoked from naiss.config -> arrhenius.config, we don't need to redo
+// everything
+
+process {
+    withName '.*' {
+        // GPUs are not available with CPU accounts, singal to modules that 
+        // they don't get any
+        accelerator = null
+    }
+
+    clusterOptions = {
+            return task.memory > 768.GB ? "-p fat" : ""
+    }
+}
+
+params.naiss_gpu_type = 'disabled'
--- a/conf/naiss/arrhenius_cpu.config
+++ b/conf/naiss/arrhenius_cpu.config
@@ -18,5 +18,5 @@ params.naiss_gpu_type = 'disabled'
 // Limit resources for 'local' if we're not in a job (if we're in a job,
 // no need to limit)
 includeConfig({
-    System.getenv("SLURM_JOB_ID") ? "limit_local.config" : "/dev/null"
+    System.getenv("SLURM_JOB_ID") ? "/dev/null" : "limit_local.config"
 }.call())
--- a/conf/naiss/arrhenius_cpu.config
+++ b/conf/naiss/arrhenius_cpu.config
@@ -3,7 +3,7 @@
 
 process {
     withName: '.*' {
-        // GPUs are not available with CPU accounts, singal to modules that 
+        // GPUs are not available with CPU accounts, signal to modules that 
         // they don't get any
         accelerator = null
     }
--- a/conf/naiss/arrhenius_cpu.config
+++ b/conf/naiss/arrhenius_cpu.config
@@ -15,7 +15,8 @@ process {
 
 params.naiss_gpu_type = 'disabled'
 
-// Limit resources for 'local' if we're not in a job
+// Limit resources for 'local' if we're not in a job (if we're in a job,
+// no need to limit)
 includeConfig({
     System.getenv("SLURM_JOB_ID") ? "limit_local.config" : "/dev/null"
 }.call())
--- a/conf/naiss/arrhenius_cpu.config
+++ b/conf/naiss/arrhenius_cpu.config
@@ -2,7 +2,7 @@
 // everything
 
 process {
-    withName '.*' {
+    withName: '.*' {
         // GPUs are not available with CPU accounts, singal to modules that 
         // they don't get any
         accelerator = null
@@ -14,3 +14,8 @@ process {
 }
 
 params.naiss_gpu_type = 'disabled'
+
+// Limit resources for 'local' if we're not in a job
+includeConfig({
+    System.getenv("SLURM_JOB_ID") ? "limit_local.config" : "/dev/null"
+}.call())
--- a/conf/naiss/arrhenius_cpu.config
+++ b/conf/naiss/arrhenius_cpu.config
@@ -9,7 +9,7 @@ process {
     }
 
     clusterOptions = {
-            return task.memory > 768.GB ? "-p fat" : ""
+        return task.memory > 768.GB ? "-p fat" : ""
     }
 }
 
--- a/conf/naiss/arrhenius_gpu.config
+++ b/conf/naiss/arrhenius_gpu.config
@@ -1,0 +1,17 @@
+
+// gpus type is only for clarity in logs and such, arrhenius only has a
+// single gpu type and we don't overcomplicate with code to break if one
+// sets another
+params.naiss_gpu_type = 'GH200'
+
+process {
+    arch = 'linux/arm64'
+
+    clusterOptions = {
+            return task.accelerator ? "-p gpu --gpus=${task.accelerator.request ?: 1}" : ""
+        }
+
+    // Enables GPU access inside the container when a task requests an accelerator.
+    containerOptions = { task.accelerator ? '--nv' : '' }
+}
+}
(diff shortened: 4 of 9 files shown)


conf/naiss.config
```diff
@@ -0,0 +1,90 @@
...
+includeConfig ({
+    def resource = "unknown"
+    try {
+        resource = ['/bin/bash', '-c', "sacctmgr show cluster -P -n | cut -f1 -d'|'"].execute().text.trim()
+    }
+    catch (IOException _e) {
+        System.err.println("WARNING: Could not run sacctmgr, defaulting to unknown")
+        return "/dev/null"
+        }
+    return resource ? "naiss/${resource}.conf": "/dev/null"
```
mahesh-panchal:
```suggestion
    return resource ? "naiss/${resource}.config": "/dev/null"
```

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+- arrhenius
+
+## Getting help
+
+We have a Slack channel dedicated to assist Swedish HPC users on the nf-core
+Slack: [https://nfcore.slack.com/channels/helpdesk-hpc-sweden](https://nfcore.slack.com/channels/helpdesk-hpc-sweden)
+
+## Using the NAISS config profile
+
+To use, run the pipeline with `-profile uppmax` (one hyphen).
```
mahesh-panchal:
```suggestion
To use, run the pipeline with `-profile naiss` (one hyphen).
```

conf/naiss/arrhenius.config
```diff
@@ -0,0 +1,38 @@
...
+// Info: https://www.naiss.se/resource/arrhenius/
+//
+
+// We only deal with cluster specific things here
+process {
+    clusterOptions = {
+        [
+
+            params.clusterOptions ?: "",
+            task.accelerator ? "-p gpu --gpus=${task.accelerator.request ?: 1}" : "",
```
mahesh-panchal: This is missing the gpu type parameter and check on task.accelerator.type

We also need to figure out how this actually works. The gpu partition I guess is only available with the GPU project allocation which likely has a different ID. 
It also means we're going to have to set `-A <account>` dynamically most likely.

This means we'll need to check if the GPU project allocation and task.accelerator are set. If task.accelerator is set and there's no GPU allocation, we're likely going to have to throw some kind of early error, preferrably before the pipeline kicks off. 

We'll need to mention to users that any GPU accelerated tools need to have container builds for ARM ( which wave should automatically do, but I'm not sure how that's flagged on modules if it doesn't ).

conf/naiss.config
```diff
@@ -0,0 +1,90 @@
...
+    clusterOptions               = null
+    schema_ignore_params         = "cluster-options,clusterOptions,project,naiss_gpu_type"
+    validationSchemaIgnoreParams = "cluster-options,clusterOptions,project,naiss_gpu_type,schema_ignore_params"
+    save_reference               = true
+   
+    // Fallback-defaults, these should be 
+    max_memory                   = 500.GB
+    max_cpus                     = 16
+    max_time                     = 240.h
+    // illumina iGenomes reference file paths on UPPMAX
```
mahesh-panchal:
```suggestion
```

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+This will download and launch the [`naiss.config`](../conf/naiss.config) which will provide general
+configuration and defer to more specific configurations per cluster that will be loaded automatically.
+
+### Cluster-speicifc configurations
```
mahesh-panchal:
```suggestion
### Cluster-specific configurations
```

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### arrhenius
+
+The configuration will submit pipeline jobs via the `Slurm` job scheduler. Tasks
+providing a container image will be run using `apptainer`. Images can be native (usually called
+`singularity` for historic reasons) or provided through `OCI`/`Docker` images.
+
+Images are downloaded and stored in a cache (usually in the `work` directory in the
+directory where you start Nextflow. If the image provided is non-native (`Docker`), it
+needds to be converted before storing. If you run out of disk space converting images set
```
mahesh-panchal:
```suggestion
needs to be converted before storing. If you run out of disk space converting images set
```

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+# Launch a nf-core pipeline with the naiss profile for the project id naiss2026-1-234
+$ nextflow run nf-core/<PIPELINE> -profile naiss --project naiss2026-1-234 [...]
+```
+
+> NB: If you're not sure what your NAISS project ID is, try running `groups` or checking SUPR.
+
+You can run `Nextflow` on a login node in a `screen` or a `tmux` session or in a job
+(batch or interactive) and it will handle everything else. The nextflow main/monitoring
+process uses very little resources.
+
```
mahesh-panchal, line 50: Add a section on GPU and add a note that GPU nodes are ARM (Grace Hopper GH200) while CPU/fat nodes are x86_64 — standard linux/amd64 nf-core containers won't run under --nv on the GPU partition unless the image has an aarch64 build.

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+Images are downloaded and stored in a cache (usually in the `work` directory in the
+directory where you start Nextflow. If the image provided is non-native (`Docker`), it
+needds to be converted before storing. If you run out of disk space converting images set
+`APPTAINER_CACHEDIR` environment variable to a location with more space.
+
+`Nextflow` also supports the environment variable `NXF_APPTAINER_CACHEDIR` which can be used to
+store and supply images for repeated executions. The equivalent `Nextflow` config setting is
+`apptainer.cacheDir`.
+
+In addition to this config profile, you will also need to specify an UPPMAX project id.
```
mahesh-panchal:
```suggestion
In addition to this config profile, you will also need to specify an NAISS project id.
```

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+### Cluster-speicifc configurations
+
+#### arrhenius
+
+The configuration will submit pipeline jobs via the `Slurm` job scheduler. Tasks
+providing a container image will be run using `apptainer`. Images can be native (usually called
+`singularity` for historic reasons) or provided through `OCI`/`Docker` images.
+
+Images are downloaded and stored in a cache (usually in the `work` directory in the
+directory where you start Nextflow. If the image provided is non-native (`Docker`), it
```
mahesh-panchal:
```suggestion
directory where you start Nextflow). If the image provided is non-native (`Docker`), it
```

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+implemented a node auto-selection system that will automatically select the
+best node for your job based on the resources you request.
+
+## Local execution
+
+For specific processes with very short runtimes, the induced latency by submitting
+the task to the scheduler and waiting for them to go through the queue may be
+non-productive. If they are light enough, you can instead have nextflow start
+them on the node where the main Nextflow process runs. The `naiss` profile
+enforces limits on such processes so they don't consume too much resources. You
```
mahesh-panchal:
```suggestion
enforces limits on such processes so they don't consume too many resources. You
```

docs/naiss.md
```diff
@@ -0,0 +1,91 @@
...
+> NB: If you're not sure what your NAISS project ID is, try running `groups` or checking SUPR.
+
+You can run `Nextflow` on a login node in a `screen` or a `tmux` session or in a job
+(batch or interactive) and it will handle everything else. The nextflow main/monitoring
+process uses very little resources.
+
+## Getting more memory
+
+If a task in your `nf-core` pipeline runs out of memory (exit code 137), you
+increase the memory request for that task by using a local config.
```
mahesh-panchal:
```suggestion
can increase the memory request for that task by using a local config.
```

docs/naiss.md
pontus: Mumble, I had forgotten/gotten confused and thought they were H200 (without the Grace). Yeah, this is a problem.

mahesh-panchal (dismissed):
Tested on Arrhenius - CPU works

```
#! /usr/bin/env bash

nextflow run nf-core/demo \
  -profile naiss,test \
  --custom_config_base https://raw.githubusercontent.com/nf-core/configs/naiss \
  -params-file params.yml \
  -resume

nextflow clean -f -before last && find work -type d -empty -delete

```

conf/naiss/arrhenius.config
pontus: The "missing" is intentional here since there's only a single one, and it didn't seem worth complicating things just to throw an error. (See https://github.com/nf-core/configs/pull/1174/changes#diff-6aa1dcfe4a72d9340b2c689c5345463e49d72f297d87fe16f3c725fb2b6b28f0R7 or up above for mentioning that in `params`).

And yes, we'll need to document, but first probably need to figure out how allocations work.

mahesh-panchal: But then just get rid of the parameter and it's extra stuff until we need it.

pontus: My thinking was that it might make sense to keep it in the general `naiss` profile, and it's included in the `arrhenius` configuration so it looks sensible in parameter logs from runs.

pontus: (So somewhat keeping the interface, but not implementing as it's not an improvement in this specific case.)

mahesh-panchal: Oh, I get it. Sure. That's useful. Keep it.

docs/naiss.md
pontus: Added a note about GPU on arrhenius. Maybe we should add a `--yes-I-really-want-to-use-gpus` as well?

conf/naiss/arrhenius.config
mahesh-panchal: Add a comment though to remind us it appears in the output and so is useful for reporting. That way I don't repeat myself the next time because I've forgotten

mahesh-panchal:
So it might be that with just a gpu project one can still run CPU pipelines as long as they have arm containers (and I guess they still go to the GPU partition). So we should probably test the project value for the suffix, and send to the GPU partition if it ends in GPU regardless if they want to use GPU or CPU. They basically want to use Arm. Then cluster options has another test on the accelerator to add the GPUs flag to slurm, and the NV flag to container options, so the GPU processes get the correct flags. If the project ends in CPU, and then I guess slurm will say it cannot allocate GPUs and error.
So groups likely only need one account.

mahesh-panchal:
Did a GPU process test from an Arrhenius CPU project. It ran without error.

```bash
#! /usr/bin/env bash

nextflow run main.nf \
  -profile naiss,test \
  --custom_config_base https://raw.githubusercontent.com/nf-core/configs/naiss \
  -params-file params.yml \
  -resume

nextflow clean -f -before last && find work -type d -empty -delete
```

```nextflow
process CHECK_GPU {
    accelerator 1
    container 'ubuntu:22.04'

    script:
    """
    echo hello from a GPU task
    nvidia-smi || echo "no nvidia-smi found"
    """

    output:
    stdout
}

workflow {
    CHECK_GPU().view()
}
```

```console
✨ Pixi task (demo-arrhenius): ./run_nextflow.sh: (Run GPU test on Arrhenius against PR #1174's naiss profile)

 N E X T F L O W   ~  version 26.04.6

WARN: It appears you have never run this project before -- Option `-resume` is ignored
Launching `main.nf` [desperate_dubinsky] revision: 0e233bb7e3

executor >  local (1)
[28/9f7d3c] CHECK_GPU [100%] 1 of 1 ✔
hello from a GPU task
Thu Sep  3 20:36:48 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.159.04             Driver Version: 580.159.04     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA L40                     On  |   00000000:82:00.0 Off |                    0 |
| N/A   39C    P8             36W /  300W |       0MiB /  46068MiB |      0%   Prohibited |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+

```

mahesh-panchal:
So then we can test the arm compatible containers too.. hmm. I should find a suitable pipeline. Rnaseq or Sarek should do.

mahesh-panchal:
I just checked the log for that GPU task. It ran locally instead of slurm. I didn't integrate the code to use the profile ...

I'll test with sarek

pontus:
Thanks for testing, although I'm not sure I'm following.

But the same proposal should be able to give access to both CPU and GPU, and my assumption would be that you only need to use one account to deal with that. One can probably do some gymnastics with `sacctmgr` to figure out what the allocation allows for - but that seems overly complex and it might also be that CPU projects are allowed GPUs, only at a reduced queue priority.

mahesh-panchal:
I think we can't use ARM CPU's only on the gpu partition.

```
ERROR ~ Error executing process > 'NFCORE_SAREK:PREPARE_GENOME:BWAMEM1_INDEX (genome.fasta)'

Caused by:
  Failed to submit process to grid scheduler for execution


Command executed:

  sbatch .command.run

Command exit status:
  1

Command output:
  sbatch: error: Your job has not allocated any GPUs.
  sbatch: error: Batch job submission failed: Invalid Trackable RESource (TRES) specification
```
`params.yml`
```yaml
outdir: "results"
aligner: "parabricks"
clusterOptions: "-p gpu"
```

Then adding the `--gpus=1` to `params.clusterOptions` results in
```

ERROR ~ Error executing process > 'NFCORE_SAREK:PREPARE_GENOME:BWAMEM1_INDEX (genome.fasta)'

Caused by:
  Failed to submit process to grid scheduler for execution


Command executed:

  sbatch .command.run

Command exit status:
  1

Command output:
  sbatch: error: Batch job submission failed: Invalid account or account/partition combination specified

```
So no gpu access from the cpu project.

pontus:
And I don't understand what went wrong with your test run, but that will be interesting. But I also think there are dragons around - e.g. I'm not sure if just running `apptainer shell docker://ubuntu` och CPU and GPU resources sequentially will work or if there is cache pollution breaking things (I've read a parts of the singularity/apptainer code base and it's /slightly/ architecture aware - here it also matters if it will call out to the registry at all or even cache that, which I don't know).

pontus:
Thanks for checking GPU access, that's good to know.

mahesh-panchal:
That's a good point and something that needs to be checked about the cache.

mahesh-panchal:
If that becomes an issue, we could try something like
```
apptainer {
    cacheDir = "${workDir}/apptainer-cache-${params.project?.endsWith('-gpu') ? 'arm64' : 'amd64'}"
}
```

mahesh-panchal:
Pavlin (who answered from NAISS support) says a user should apply for both a GPU and CPU project if they want to use both. cpu and fat are exclusive to CPU project. gpu is exclusive to GPU. 

The project id should have the same stem `naiss-XXX-{cpu,gpu}`.

pontus:
Ok, hohum. We should disable GPU if a `cpu` account is used at least, I guess then (easy enough to do if it's in the name, if that `withName` works).

Then I wonder what's the best strategy for dealing with GPUs, but at least we won't have a mix in the same run.

mahesh-panchal:
I think a mix is what they're expecting us to do

mahesh-panchal:
One thing Pavlin suggested was that for a GPU project, it might be better to book a whole node and run locally on the node. That way we can take advantage of all the ARM CPU's that way. If we're submitting via slurm, we'll likely make splits of 1 GPU + 72 CPUs + whatever memory ( and likely overbook because of memory - my take ).

conf/naiss/arrhenius.config
```diff
@@ -0,0 +1,33 @@
...
+        "arrhenius_cpu.config" :
+        "/dev/null"
+    }.call())
+
+// and GPU, we only do these for explicit account names
+includeConfig({
+    (params.account && params.account.trim().endsWith("-gpu")) ?
+        "arrhenius_gpu.config" :
+        "/dev/null"
+    }.call())
```
mahesh-panchal, line 25: Why not a single include since it's on params.account?

I think the split is a good idea though. Let's keep it either-or for now.

conf/naiss/arrhenius_cpu.config
```diff
@@ -0,0 +1,16 @@
...
+
+process {
+    withName '.*' {
+        // GPUs are not available with CPU accounts, singal to modules that 
+        // they don't get any
+        accelerator = null
+    }
+
+    clusterOptions = {
+            return task.memory > 768.GB ? "-p fat" : ""
```
mahesh-panchal:
```suggestion
        return task.memory > 768.GB ? "-p fat" : ""
```

conf/naiss/arrhenius_gpu.config
```diff
@@ -0,0 +1,17 @@
...
+process {
+    arch = 'linux/arm64'
+
+    clusterOptions = {
+            return task.accelerator ? "-p gpu --gpus=${task.accelerator.request ?: 1}" : ""
+        }
+
+    // Enables GPU access inside the container when a task requests an accelerator.
+    containerOptions = { task.accelerator ? '--nv' : '' }
+}
```
mahesh-panchal, line 21: I think there's an extra bracket and formatting issues here.

conf/naiss.config
```diff
@@ -0,0 +1,89 @@
...
+executor {
+    $slurm {
+        account = params.project
+    }
+
+    // Cap local jobs to limit burden on the node they run on
+    $local {
+        cpus   = 8
+        memory = 128.GB
+    }
```
mahesh-panchal: I think we should move this on to the CPU config.

docs/naiss.md
```diff
@@ -0,0 +1,118 @@
...
+So far, we haven't come up with a good solution for this. In the meantime,
+you can tell nextflow to use specific container images for the tasks requiring
+GPU resources.
+
+Recent releases of nf-core pipelines should have a file
+`conf/containers_singularity_https_arm64.config` which you can use to find the
+Nextflow processes you need to run on GPU resources and copy those clauses to
+a file `nextflow.config` in the directory you run nextflow from (which will make
+it used without you asking, see
+[Seqera documentation](https://docs.seqera.io/nextflow/config)) for other options.
```
mahesh-panchal: They shouldn't have to copy these files anywhere, but I think it's good to know that pipelines should have this in their code if they support arm.

docs/naiss.md
```diff
@@ -0,0 +1,118 @@
...
+
+But from e.g. https://github.com/nf-core/funcprofiler/blob/dev/conf/containers_singularity_https_arm64.config, you can
+copy the line
+
+```nextflow
+process { withName: 'MULTIQC' { container = 'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/e4/e48aa28aebc881254a499b24c3e1ce77b8df1b85a5432699ed6f72eb17ac7fb5/data' } }
+```
```
mahesh-panchal: Rather than copy across, if a profile is not autoloading that config, they should simply using an `includeConfig` statement themselves.

docs/naiss.md
```diff
@@ -0,0 +1,118 @@
...
+best node for your job based on the resources you request.
+
+## Local execution
+
+For specific processes with very short runtimes, the induced latency by submitting
+the task to the scheduler and waiting for them to go through the queue may be
+non-productive. If they are light enough, you can instead have nextflow start
+them on the node where the main Nextflow process runs. The `naiss` profile
+enforces limits on such processes so they don't consume too many resources. You
+should still be aware and restrictive with local execution.
```
mahesh-panchal, line 98: You can suggest the `maxForks` directive, but the config code `executor.$local` should limit how much runs anyway.

mahesh-panchal (commented):
I like the split. I think for running on a GPU project, it's probably better that they just book a whole node and run locally. That way they can make use of the CPU's even if they don't use GPU.
We would just need to make a check somehow that they're not running on the login node since they would use the local executor rather than slurm. Maybe check that a SLURM env has a variable. If not, stop.

docs/naiss.md
pontus: This was for mixed runs (x86_64/arm64), if we don't do that, it can go.

conf/naiss/arrhenius.config
pontus: For now, I thought it was good to deal with other account names, I'm not sure if it's super useful but I started that way here.

conf/naiss/arrhenius_gpu.config
pontus: Correct, fixed.

conf/naiss/arrhenius_cpu.config
pontus: Fixed.

docs/naiss.md
pontus: This was for mixed runs (x86_64/arm64), if we don't do that, it can go.

conf/naiss/arrhenius.config
mahesh-panchal: I don't understand. There's only one `params.account`, or did you meant to make two parameters to support using both partitions simultaneously?

mahesh-panchal:
What are your thoughts on only supporting one account at a time? My thinking is that the GPU allocations are horribly defined that using the slurm allocation per GPU is going to be expensive in comparison to running the whole pipeline locally on one node.

conf/naiss/arrhenius.config
pontus: I don't have enough info yet but haven't heard that a single account can't have cpu and gpu allocations simultaneously even if the naming slightly suggests so. Or neither.

For now, I'm happy being very explicit about testing.

conf/naiss.config
pontus: I'm currently thinking it can be gated on whatever we're not in a slurm job (some appropriate environment variable not set).

pontus:
I'm not sure I agree about GPU allocations but think we can/should start with not caring about mixed runs. I think I can come up with something to deal with local runs.

conf/naiss/arrhenius_cpu.config
```diff
@@ -0,0 +1,22 @@
...
+        return task.memory > 768.GB ? "-p fat" : ""
+    }
+}
+
+params.naiss_gpu_type = 'disabled'
+
+// Limit resources for 'local' if we're not in a job (if we're in a job,
+// no need to limit)
+includeConfig({
+    System.getenv("SLURM_JOB_ID") ? "limit_local.config" : "/dev/null"
```
mahesh-panchal: Shouldn't this be the other way around? If SLURM_JOB_ID -> /dev/null.

conf/naiss/arrhenius_cpu.config
```diff
@@ -0,0 +1,22 @@
+// Invoked from naiss.config -> arrhenius.config, we don't need to redo
+// everything
+
+process {
+    withName: '.*' {
+        // GPUs are not available with CPU accounts, singal to modules that
```
mahesh-panchal:
```suggestion
        // GPUs are not available with CPU accounts, signal to modules that 
```

conf/naiss/arrhenius.config
mahesh-panchal: Also it's `params.project` not `params.account`.

docs/naiss.md
```diff
@@ -0,0 +1,118 @@
...
+
+> NB: If you're not sure what your NAISS project ID is, try running `groups` or checking SUPR.
+
+You can run `Nextflow` on a login node in a `screen` or a `tmux` session or in a job
+(batch or interactive) and it will handle everything else. The nextflow main/monitoring
+process uses very little resources.
+
+##### GPU on arrhenius
+
+GPUs are offerend on arrhenius through GH200 "superchip" nodes, which uses
```
mahesh-panchal:
```suggestion
GPUs are offered on arrhenius through GH200 "superchip" nodes, which uses
```

conf/naiss/arrhenius.config
```diff
@@ -0,0 +1,33 @@
...
+    config_profile_description = 'NAISS Arrhenius cluster profile provided by nf-core/configs.'
+    
+    max_cpus                   = 192
+    max_memory                 = 3.TB
+    max_time                   = 3.d
+}
+
+// Some gymnastics to get conditional inclusion, here to cover cpu only
+includeConfig({
+    (params.account && params.account.trim().endsWith("-cpu")) ?
```
mahesh-panchal:
```suggestion
    (params.project && params.project.trim().endsWith("-cpu")) ?
```

conf/naiss/arrhenius.config
```diff
@@ -0,0 +1,33 @@
...
+        "arrhenius_cpu.config" :
+        "/dev/null"
+    }.call())
+
+// and GPU, we only do these for explicit account names
+includeConfig({
+    (params.account && params.account.trim().endsWith("-gpu")) ?
```
mahesh-panchal:
```suggestion
    (params.project && params.project.trim().endsWith("-gpu")) ?
```

conf/naiss/arrhenius_cpu.config
```diff
@@ -0,0 +1,22 @@
...
+        return task.memory > 768.GB ? "-p fat" : ""
+    }
+}
+
+params.naiss_gpu_type = 'disabled'
+
+// Limit resources for 'local' if we're not in a job (if we're in a job,
+// no need to limit)
+includeConfig({
+    System.getenv("SLURM_JOB_ID") ? "limit_local.config" : "/dev/null"
```
mahesh-panchal:
```suggestion
    System.getenv("SLURM_JOB_ID") ? "/dev/null" : "limit_local.config"
```

docs/naiss.md
pontus: This is mostly a general caution to be considerate (and e.g. creating millions of files in `/tmp` or doing other stuff that's likely to affect others (that nextflow limits won't manage).

conf/naiss/arrhenius.config
pontus: 🤦

docs/naiss.md
pontus: Updated.

pontus:
I think this is getting good enough to merge and then if people use it and report issues we'll deal with it then, does that sound good or does anyone see something that needs improvement?

mahesh-panchal:
**Resource limits vs. actual hardware** ([source](https://www.naiss.se/resources/arrhenius-technical-description/)):

- `max_cpus = 192` is below every node's real core count (CPU nodes: 256, GPU nodes: 288). Intentional, or placeholder?
- `max_memory = 3.TB` is correct for CPU/fat, but `arrhenius_gpu.config` inherits it unchanged — GPU nodes have ~512 GB, not 3 TB.

pontus:
Thanks, I managed to put the number of nodes as `max_cpus`. :facepalm: 

For GPUs, those are set in the GPU config (although the values from slurm as reported to me give just over 400 Gbyte).

mahesh-panchal (approved):

PR #1175, opened 2026-09-04
Remove 'scratch' option from executor Cambridge configuration

RaqManzano:
Removed the 'scratch' option from the executor configuration. We have discovered though usage that `scratch` option is not working well for our cluster as the default `tmp` gets full quite easily. Therefore I am removing this option from the config, users can still activate it through their config if they wish to.

### DIFF
--- a/conf/cambridge.config
+++ b/conf/cambridge.config
@@ -69,7 +69,6 @@ process {
     queue          = params.partition
     clusterOptions = { params.project ? "--account=${params.project}" : '' }
     cache          = 'lenient'
-    scratch        = true
 }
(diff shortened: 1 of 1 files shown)


RaqManzano:
I am getting some failed checks. Could someone let me know if these are necessary for the PR? I believe the failed checks are unrelated to the Cambridge config.

pontus (approved):

RaqManzano:
Hi @pontus, I am unable to merge even though I see you have approved the PR. Are you able to do it on your end? Thanks.

